PR #4, opened 2017-08-04
A-2: I want to connect Open edX as a Content Source via LTI protocol

wowkalucky:
Implements simple view to make a connection from the Bridge as LTI consumer to the OpenEdx instance through LTI protocol (using pylti lib).

Also, code style fixes are made in modules which were edited.
Updated secure.py.example file to make the format of django-auth-lti settings clearer.
Implemented creation of Collections, Activities and Sources (aka LTI content sources).
Added basic user login/logout flow.


idegtiarov:
First of all, provide please description for the PR

bridge-adaptivity/bridge_lti/views.py
```diff
@@ -48,3 +77,35 @@ def launch(request):
     return redirect('module:launch', user_module_id=user_module.pk)
 
 
+def content_source(request):
+    """
+    Simple view to render Source's component shared through LTI.
+    """
+    consumer = ToolConsumer(
+        # STAGE: hardcoded values
```
kunanit: Do you plan to remove these values later?

wowkalucky: sure!

PR #6, opened 2017-08-10
I10.1 - As an administrator I want to use a bridge as lti tool.

wowkalucky:
Implemented naive LTI provider launch view.
Performed minor markup fixes.

Results: Bridge Instructor interface is available through LTI OpenEdx xblock studio view.


wowkalucky:
Changed branch name to user-prefixed.

PR #11, opened 2017-08-21
Create adaptive engine example

kunanit:
Example code for the adaptive engine

warning - probably not functional yet (don't merge yet)


kunanit:
A full adaptive engine implementation example can be seen at https://github.com/harvard-vpal/adaptive-engine

PR #14, opened 2017-08-22
I12: Activity configuration

wowkalucky:
- removed temporary LtiSource model (source related data moved to Activity);
- removed ContentSources list and detail view;
- refactored content_source view;
- added modal Activity creation form;
- added modal source blocks preview;
- added content source's courses list fetching (API);
- added content source's blocks fetching (API);
- fixed template markup;
- implemented OAuth client for OpenEdx API connection;
- updated lti launch routing;

PR #15, opened 2017-08-23
Data schema design fixes

wowkalucky:
Note: branched from `dev`. So it should wait for #14.

Changes:
- added `threshold` field to Collection model;
- mutated Activity model `difficulty` field from float to 3-leveled choices;
- changed Activity model `tag` field to `tags`;


bridge-adaptivity/module/models.py
```diff
@@ -66,10 +68,16 @@ class Activity(models.Model):
...
+        ('medium', _('medium')),
+        ('high', _('high')),
+    )
+
     name = models.CharField(max_length=255)
     collection = models.ForeignKey('Collection', null=True)
-    tag = fields.CharField(max_length=255, blank=True, null=True)
-    difficulty = models.FloatField(verbose_name="max_points", blank=True, null=True)
+    tags = fields.CharField(max_length=255, blank=True, null=True)
+    difficulty = fields.CharField(choices=LEVELS, default='medium', max_length=32)
```
idegtiarov: I would use 'l', 'm', 'h' keys for LEVEL and decrease difficulty field's max-length to 1 character

PR #16, opened 2017-08-28
Update logging to be more informative.

wowkalucky:
- Make debug messages more informative while OAuth OpenEdx Provider is disabled.
- Minor exception catching refactoring.
- Squash (recreate) Module application's migrations (had data merge problems on the stage).
- Improve templates informativity for bridge instructors.
- Bugfix: remove renamed field old name usage (blocker).


bridge-adaptivity/api/backends/openedx.py
```diff
@@ -60,8 +62,11 @@ def get_oauth_access_token(self):
             log.exception(
                 'OAuth2 token request to the OpenEdx LTI Provider failed: {}'.format(exc.message)
             )
+            log.warn("You may want to check your OAuth registration on LTI Provider."
```
idegtiarov: Why do we need two logging messages if the exception is raised one well-informed message would be fair enough.

bridge-adaptivity/api/backends/openedx.py
```diff
@@ -137,8 +142,8 @@ def get_available_blocks(course_id):
         )
     except HttpNotFoundError:
         raise HttpClientError(_("Requested course not found. Check `course_id` url encoding."))
-    except HttpClientError:
-        raise HttpClientError(_("Not valid query."))
+    except HttpClientError as exc:
+        raise HttpClientError(_("Not valid query: {original}").format(original=exc.message))
```
idegtiarov: Nit: you could just use "Not valid query: {}").format(exc.message)

bridge-adaptivity/api/backends/openedx.py
```diff
@@ -159,8 +164,8 @@ def get_available_courses():
     try:
         courses_list = api.get_provider_courses()
         filtered_courses = apply_data_filter(courses_list, filters=['id', 'course_id', 'name', 'org'])
-    except HttpClientError:
-        raise HttpClientError(_("Not valid query."))
+    except HttpClientError as exc:
+        raise HttpClientError(_("Not valid query: {original}").format(original=exc.message))
```
idegtiarov: ditto

bridge-adaptivity/module/views.py
```diff
@@ -52,15 +52,12 @@ def get_context_data(self, **kwargs):
     def get_content_courses():
         try:
             return get_available_courses()
-        except ObjectDoesNotExist as exc:
+        except HttpClientError as exc:
             log.error(
```
idegtiarov: Use exception type of the logging message and exc.message is not required to be added in that case.

bridge-adaptivity/api/backends/openedx.py
```diff
@@ -60,8 +62,11 @@ def get_oauth_access_token(self):
             log.exception(
                 'OAuth2 token request to the OpenEdx LTI Provider failed: {}'.format(exc.message)
```
idegtiarov: exc.message is not needed with exception type of the logging message

bridge-adaptivity/docker-compose.yml
```diff
@@ -27,7 +27,7 @@ services:
     container_name: nginx_BFA
     build: ./nginx
     ports:
-      - "80:80"
+      - "8080:80"
```
wowkalucky: This is because of our stage configuration Gamanenko's quick hack: we have system nginx which listens 80 and proxies all to 8080.

idegtiarov:
Thanks for fast responce.

PR #17, opened 2017-08-28
Collection editing.

wowkalucky:
- Implement Activity deletion.


bridge-adaptivity/module/urls.py
```diff
@@ -16,7 +18,13 @@
         login_required(ActivityUpdate.as_view()),
         name='activity-change'
     ),
+    url(
+        r'^activity/(?P<pk>\d+)/delete/$',
+        login_required(ActivityDelete.as_view()),
```
idegtiarov: I would keep view logic outside from the URL config.

bridge-adaptivity/module/urls.py
```diff
@@ -16,7 +18,13 @@
         login_required(ActivityUpdate.as_view()),
```
idegtiarov: I would keep view logic outside from the URL config.

PR #19, opened 2017-08-29
Make trailing slash in Bridge LTI launch URLs optional

wowkalucky:
- Made LTI launch URL to catch:
-- lti/launch;
-- lti/launch/;
-- lti/launch/1;
-- lti/launch/1/;
- Refactor Instructor to be redirected to:
-- collections list view by the base lti url (w/o collection_id);
-- specific collection detail view by the lti url with provided collection_id.

### DIFF
--- a/bridge-adaptivity/bridge_lti/provider.py
+++ b/bridge-adaptivity/bridge_lti/provider.py
@@ -68,4 +68,3 @@ def lti_launch(request, collection_id=None):
         sequence_item = sequence.sequenceitem_set.last()
 
     return redirect(reverse('module:sequence-item', kwargs={'pk': sequence_item.id}))
-
--- a/bridge-adaptivity/bridge_lti/provider.py
+++ b/bridge-adaptivity/bridge_lti/provider.py
@@ -53,15 +53,29 @@ def lti_launch(request, collection_id=None):
                 'tip': 'have a look at `consumer secret`',
             }
         )
-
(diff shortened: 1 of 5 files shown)


bridge-adaptivity/bridge_lti/provider.py
```diff
@@ -54,45 +54,50 @@ def lti_launch(request, collection_id=None):
...
+                'tip': 'this adaptivity sequence is about to start.',
+            }
         )
 
-        if sequence.completed:
-            return redirect(reverse('module:sequence-complete', kwargs={'pk': sequence.id}))
+    try:
+        collection = Collection.objects.get(id=collection_id)
+    except Collection.DoesNotExist:
+        return HttpResponseBadRequest()
```
idegtiarov: I would add some logs here with the description of the error.

bridge-adaptivity/bridge_lti/urls.py
```diff
@@ -5,7 +5,6 @@
 from .consumer import source_preview
 
 urlpatterns = [
-    url(r'^launch$', lti_launch, name='initial-launch'),
-    url(r'^launch/(?P<collection_id>\d+)$', lti_launch, name='launch'),
+    url(r'^launch/?(/(?P<collection_id>\d+))?/?$', lti_launch, name='launch'),
```
idegtiarov: This pattern looks a little bit ugly: why don't use something like this: `r'^launch/(?P<collection_id>\d*)$'` ?

PR #22, opened 2017-08-30
Make informative feedback for LMS while LTI misconfigurations

wowkalucky:
If wrong LTI credentials are provided by Instructor during LMS LTI tool configuration the informative message is shown.

### DIFF
--- a/bridge-adaptivity/bridge_lti/provider.py
+++ b/bridge-adaptivity/bridge_lti/provider.py
@@ -1,4 +1,4 @@
-from django.http import HttpResponseBadRequest, HttpResponseForbidden
+from django.http import HttpResponseBadRequest
 from django.shortcuts import redirect, render
 from django.urls import reverse
 from django.views.decorators.csrf import csrf_exempt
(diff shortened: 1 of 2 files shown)


bridge-adaptivity/bridge_lti/provider.py
```diff
@@ -26,14 +26,40 @@ def lti_launch(request, collection_id=None):
     try:
         lti_consumer = LtiProvider.objects.get(consumer_key=params['oauth_consumer_key'])
     except LtiProvider.DoesNotExist:
-        return HttpResponseForbidden()
+        # wrong 'consumer_key':
```
idegtiarov: I would add some exception logs here

bridge-adaptivity/bridge_lti/provider.py
```diff
@@ -26,14 +26,40 @@ def lti_launch(request, collection_id=None):
...
+            context={
+                'title': 'forbidden',
+                'message': 'please, check provided LTI credentials.',
+                'tip': 'have a look at `consumer key`',
+            }
+        )
 
     if not SignatureValidator(lti_consumer).verify(request):
-        return HttpResponseForbidden()
+        # wrong 'consumer_secret':
```
idegtiarov: ditto

bridge-adaptivity/bridge_lti/provider.py
```diff
@@ -26,14 +26,40 @@ def lti_launch(request, collection_id=None):
...
+            request,
+            template_name="bridge_lti/announcement.html",
+            context={
+                'title': 'forbidden',
+                'message': 'please, check provided LTI credentials.',
+                'tip': 'have a look at `consumer secret`',
+            }
+        )
 
     if params['roles'] in ['Student', 'Learner']:
```
idegtiarov: This statement should be rewritten params['roles'] returns string with the roles separated with the comma so you could get `params['roles'] == 'Instructor,Administrator, Mentor'` 

Please take a look at docs: https://www.imsglobal.org/specs/ltiv1p1/implementation-guide#toc-11

And change statement to the opposite check if role is Instructor otherwise all roles are proposed student view

wowkalucky: This will be addressed in the #19

bridge-adaptivity/bridge_lti/provider.py
```diff
@@ -27,6 +30,7 @@ def lti_launch(request, collection_id=None):
         lti_consumer = LtiProvider.objects.get(consumer_key=params['oauth_consumer_key'])
     except LtiProvider.DoesNotExist:
         # wrong 'consumer_key':
+        log.warn('LTI: provided wrong consumer key.')
```
idegtiarov: It would be better to use exception type of the logs in the try/except block

bridge-adaptivity/bridge_lti/provider.py
idegtiarov: Got it thanks

PR #26, opened 2017-09-04
After project refactoring merge fixes

wowkalucky:
Configuration fixes missed during project structure refactoring.


bridge_adaptivity/bridge_lti/urls.py
```diff
@@ -4,6 +4,6 @@
 from .consumer import source_preview
 
 urlpatterns = [
-    url(r'^launch/(?P<collection_id>\d*)$', lti_launch, name='launch'),
```
wowkalucky: Extra one...

bridge_adaptivity/bridge_lti/urls.py
```diff
@@ -4,6 +4,6 @@
...
+    url(r'^launch(?P<collection_id>\d*)$', lti_launch, name='launch'),
```
idegtiarov: That is not correct, slash is required here, please revert.

wowkalucky: But it doesn't catch the `/lti/launch` case.

idegtiarov: Why we need to care about '/lti/launch' when we have `/lti/launch/`?

wowkalucky: Because of 50% of cases, instructors will mistyped it?

idegtiarov: That is not the case, django appends the slash to the end of the URL by default.

PR #29, opened 2017-09-06
Add inline activity preview

wowkalucky:
- updated debug logs;
- updated docker-compose dev config (to be able to attach to the service for debugging);
- updated templates;
- added Activity inline options: preview, delete;
- restricted API source components to 'video', 'html', 'problem' types;
- added problem type recognition to UI;

### DIFF
--- a/bridge_adaptivity/bridge_lti/consumer.py
+++ b/bridge_adaptivity/bridge_lti/consumer.py
@@ -1,3 +1,5 @@
+import logging
+
 from django.contrib.auth.decorators import login_required
 from django.http import HttpResponse
 from django.shortcuts import render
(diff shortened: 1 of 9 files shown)


bridge_adaptivity/bridge_lti/consumer.py
```diff
@@ -32,6 +36,9 @@ def source_preview(request):
     """
     Simple view to render Source content block shared through LTI.
     """
+    log.debug("Got request.POST: %s", request.POST)
```
idegtiarov: It seems that we are expecting only GET request for this view, don't we?

idegtiarov:
JS Should be refactored and moved into separate file

PR #30, opened 2017-09-07
Implement policy-1a grade calculation

wowkalucky:
- Sequence got new field 'trials' - number of ever submitted activities;
- Altered Collection.threshold to PositiveIntegerField;
- moved Sequence.total_points to SequenceItem.points;
- added new utility to calculate grade (using grade policy '1a');
- updated docker-compose config;


bridge_adaptivity/bridge_lti/provider.py
```diff
@@ -67,9 +67,9 @@ def lti_launch(request, collection_id=None):
         return learner_flow(request, lti_consumer, params, collection_id=collection_id)
 
 
-def instructor_flow(request, collection_id=None):
+def instructor_flow(_request, collection_id=None):
```
idegtiarov: Why do you start function param name with an underscore? It looks odd

wowkalucky: Tried to stress that var isn't used. There is convention about '_', but in Django it is used as gettext.
So in Edx I saw '__' usage. Will it work?

idegtiarov: If the parameter is not used why do we need it in the function?

PR #33, opened 2017-09-13
Add LTI session which is used for student views

idegtiarov:
Add LTI session which is used for student views
Improve response xml for grade callback function
Add checking for LTI session breaks which lead to updating activity in un-submitted sequence item

### DIFF
--- a/bridge_adaptivity/config/settings/local.py
+++ b/bridge_adaptivity/config/settings/local.py
@@ -5,12 +5,12 @@ os.environ.setdefault('ENV_TYPE', 'local')
 
 DEBUG = True
 
-INSTALLED_APPS += ('debug_toolbar', 'sslserver')
-
-MIDDLEWARE_CLASSES += (
-    'djdev_panel.middleware.DebugMiddleware',
-    'debug_toolbar.middleware.DebugToolbarMiddleware',
-)
+# INSTALLED_APPS += ('debug_toolbar', 'sslserver')
+#
+# MIDDLEWARE_CLASSES += (
+#     'djdev_panel.middleware.DebugMiddleware',
+#     'debug_toolbar.middleware.DebugToolbarMiddleware',
+# )
 
 STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.StaticFilesStorage'
 STATIC_URL = '/static/'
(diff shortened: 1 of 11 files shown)


bridge_adaptivity/config/settings/local.py
```diff
@@ -5,12 +5,12 @@
...
-MIDDLEWARE_CLASSES += (
-    'djdev_panel.middleware.DebugMiddleware',
-    'debug_toolbar.middleware.DebugToolbarMiddleware',
-)
+# INSTALLED_APPS += ('debug_toolbar', 'sslserver')
+#
+# MIDDLEWARE_CLASSES += (
+#     'djdev_panel.middleware.DebugMiddleware',
+#     'debug_toolbar.middleware.DebugToolbarMiddleware',
+# )
```
wowkalucky: git update-index --assume-unchanged local.py

bridge_adaptivity/module/utils.py
```diff
@@ -59,4 +92,19 @@ def parse_callback_grade_xml(body):
...
         raise BridgeError('score value outside the permitted range of 0.0-1.0')
 
-    return sourced_id, score
+    return imsx_message_identifier, sourced_id, score, action
+
+
+def chose_activity(sequence_item, sequence=None):
+    if sequence_item is None:
+        sequence = sequence
+    else:
```
wowkalucky: Do we really need increase complexity here?

bridge_adaptivity/module/views.py
```diff
@@ -146,33 +150,34 @@ def sequence_item_next(request, pk):
...
-        next_sequence_item = SequenceItem.objects.create(
-            sequence=sequence_item.sequence,
-            activity=activity,
-            position=sequence_item.position + 1
-        )
-
+    log.debug("Picked next sequence item is: {}".format(next_sequence_item))
+
+    if not next_sequence_item or next_sequence_item.position == last_item:
+        activity = utils.chose_activity(sequence_item)
```
wowkalucky: It looks like this `if` can be inverted for the sake of sanity...

bridge_adaptivity/config/settings/local.py
idegtiarov: 👍

bridge_adaptivity/module/utils.py
idegtiarov: Actually, we have no choice and have to decide which sequence should be used, but I will rewrite this.

bridge_adaptivity/module/views.py
idegtiarov: sure, nice catch!

PR #35, opened 2017-09-14
Move engine update grade entry point to SequenceItem save() method

idegtiarov:
Add CI's tabs into README
Fix timeout param in the nonce cached instance


bridge_adaptivity/module/models.py
```diff
@@ -40,6 +40,12 @@ class SequenceItem(models.Model):
     position = models.PositiveIntegerField(default=1)
     score = models.FloatField(null=True, blank=True, help_text="Grade policy: 'p' (problem's current score).")
 
+    __origin_score = None
```
wowkalucky: Just curious, where it will be used outside?...

PR #39, opened 2017-09-18
Implement log type 'admin'

wowkalucky:
- Added new fields to Log model: 'action', 'data';
- Extended Log types with 'Admin' type;
- Added 'Admin' type logs creation while Action/Collection instance creates/updates/deletes;
- quick fix: collection create form button style.


bridge_adaptivity/module/models.py
```diff
@@ -85,6 +86,24 @@ class Meta:
     def __str__(self):
         return u'<Collection: {}>'.format(self.name)
 
+    def save(self, *args, **kwargs):
+        """
+        Extend save() method with logging.
+        """
+        initial_id = self.id
+        super(Collection, self).save(*args, **kwargs)
+
```
wowkalucky: Here we have 'create' + 'update' logs during Activity creation...

bridge_adaptivity/module/models.py
```diff
@@ -142,19 +161,36 @@ def save(self, *args, **kwargs):
         """
         Extend save() method with sending notification to the Adaptive engine that Activity is created/updated
         """
-        if Activity.objects.filter(id=self.id):
+        initial_id = self.id
+        super(Activity, self).save(*args, **kwargs)
+
+        if initial_id:
```
idegtiarov: 👍

PR #41, opened 2017-09-18
Restrict activity order option buttons

wowkalucky:
- Do not show order buttons for not actual Activity movements.


bridge_adaptivity/module/templates/module/collection_detail.html
```diff
@@ -57,14 +57,20 @@
...
                     <a class="move-up pull-right"
                        href="{% url 'module:activity-move' pk=activity.pk direction='up' %}">
                       {% bootstrap_button "" size='sm' icon='chevron-up' %}
                     </a>
-                    <a class="move-down pull-right"
-                       href="{% url 'module:activity-move' pk=activity.pk direction='down' %}">
-                      {% bootstrap_button "" size='sm' icon='chevron-down' %}
-                    </a>
+                    {% endifnotequal %}
+                    {% if not forloop.last %}
```
idegtiarov: Why do we check last in forloop and lastpre in different condition statements you can use one with the OR operator.

wowkalucky: actually I tried... It didn't work...

PR #42, opened 2017-09-18
Add collection update view.

wowkalucky:
- Added Collection update view;
- Added options edit button to Collection list view;


bridge_adaptivity/module/templates/module/collection_list.html
```diff
@@ -25,6 +26,14 @@
             <span class="glyphicon glyphicon-minus-sign" aria-hidden="true"></span>
           {% endif %}
           </td>
+          <td>
+            <div>
+              <a class="move-down pull-right"
```
idegtiarov: I would use default behavior with left alignment like it is done to the strict forward marker

PR #43, opened 2017-09-19
Add launch URL fetch button

wowkalucky:
speaks for itself


bridge_adaptivity/module/templates/module/base.html
```diff
@@ -11,6 +11,7 @@
   <title>{% block bootstrap3_title %}Bridge DASHBOARD{% endblock %}</title>
   {% bootstrap_css %}
   <link rel="stylesheet" href="{% static 'module/css/module.css' %}"/>
+  <script src="{% static 'module/js/vendor/clipboard.min.js' %}"></script>
```
wowkalucky: It is possible to use one of remote [storages](https://github.com/zenorocha/clipboard.js/wiki/CDN-Providers), but I experienced a page loading delay ones, so I decided include script to static...

bridge_adaptivity/module/views.py
```diff
@@ -16,10 +16,10 @@
 from api.backends.openedx import get_available_courses, get_content_provider
 from bridge_lti import outcomes
 from bridge_lti.outcomes import calculate_grade
+from module import utils
 from module.forms import ActivityForm
 from module.mixins import CollectionIdToContextMixin, LtiSessionMixin
 from module.models import Collection, Activity, SequenceItem, Log, Sequence
-from module import utils
```
wowkalucky: Made imports `true`

idegtiarov: 👍

bridge_adaptivity/module/templates/module/base.html
idegtiarov: That is really not needed, CDN link would be fair enough. After page loading js lib will be cached.

PR #64, opened 2017-10-12
Workaround for when session variables are not available for iframe

kunanit:
Depending on user's browser settings, the LTI session variables cannot be found - likely due to the problem described in [this link](http://www.mendoweb.be/blog/internet-explorer-safari-third-party-cookie-problem/)

One workaround is to ask the user to visit a page on the iframe host (in this case the bridge). This PR adds a view at /lti/create_session to create a new session and set the cookie manually. After the learner visits that page in a new tab or window, they are able to use the bridge normally.

It also seems like after the learner is able to access the bridge (and get content served from the openedx content source), some javascript from the content source is not able to load, and the submit button in the problem becomes disabled. A similar solution where the learner visits a page on the content source fixes this problem.

This PR adds a message to the learner when the LTI session is not found, that directs them to open pages on the content source and the bridge in new tab/window, close them, then refresh the LMS page in order to start the adaptive quiz.

### DIFF
--- a/bridge_adaptivity/bridge_lti/provider.py
+++ b/bridge_adaptivity/bridge_lti/provider.py
@@ -1,7 +1,7 @@
 import logging
 
 from django.core.cache import cache
-from django.http import Http404, HttpResponseBadRequest
+from django.http import Http404, HttpResponseBadRequest, HttpResponse
 from django.shortcuts import redirect, render
 from django.urls import reverse
 from django.views.decorators.csrf import csrf_exempt
@@ -145,3 +145,8 @@ def learner_flow(request, lti_consumer, tool_provider, collection_id=None):
         return redirect(reverse('module:sequence-complete', kwargs={'pk': sequence.id}))
 
     return redirect(reverse('module:sequence-item', kwargs={'pk': sequence_item_id}))
+
+
+def create_session(request):
+    request.session.cycle_key()
+    return HttpResponse()
--- a/bridge_adaptivity/bridge_lti/provider.py
+++ b/bridge_adaptivity/bridge_lti/provider.py
@@ -148,5 +148,8 @@ def learner_flow(request, lti_consumer, tool_provider, collection_id=None):
 
 
 def create_session(request):
+    """
+    View that creates a new django session
+    """
     request.session.cycle_key()
-    return HttpResponse()
+    return HttpResponse("New user session created")
--- a/bridge_adaptivity/bridge_lti/urls.py
+++ b/bridge_adaptivity/bridge_lti/urls.py
@@ -1,9 +1,10 @@
 from django.conf.urls import url
 
 from bridge_lti.consumer import source_preview
-from bridge_lti.provider import lti_launch
+from bridge_lti.provider import lti_launch, create_session
 
 urlpatterns = [
     url(r'^launch/?(?P<collection_id>\d*)/?$', lti_launch, name='launch'),
     url(r'^source/$', source_preview, name='source-preview'),
+    url(r'^create_session/$', create_session),
 ]
--- a/bridge_adaptivity/module/mixins.py
+++ b/bridge_adaptivity/module/mixins.py
@@ -32,10 +32,12 @@ class LtiSessionMixin(object):
         if not lti_session:
             log.error('Lti session is not found, Request cannot be processed')
             msg = """
-Your browser security settings require an additional step before starting the quiz.
-Please open https://courses.openedx.vpal.io/event and {}/lti/create_session in new browser tabs.
-Then close those tabs, and refresh this page.
-""".format(settings.BRIDGE_HOST)
+                Your browser security settings require an additional step before starting the quiz.
+                Please open https://courses.openedx.vpal.io/event and {}/lti/create_session in new
+                browser tabs. Then close those tabs, and refresh this page. If you continue experiencing
+                problems accessing the quiz, try decreasing your browser's security settings or using a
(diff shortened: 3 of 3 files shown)


kunanit:
The original issue could be tricky to reproduce since its related to user browser settings, but what worked for me is to use the Safari browser, clear the browser history/cookies, then try to use a Bridge LTI component in the LMS.

idegtiarov:
Nice catch Andrew! I'm not sure we need any new endpoint for users to create new cookies for the bridge site. I would extend the error message to cover this issue with the proposition to decrease security browser level to start working on the adaptive unit and after that increase, it back, or open the new tab with the bridge's index page and close it after that unit will be rendered correctly.

bridge_adaptivity/bridge_lti/provider.py
```diff
@@ -145,3 +145,11 @@ def learner_flow(request, lti_consumer, tool_provider, collection_id=None):
         return redirect(reverse('module:sequence-complete', kwargs={'pk': sequence.id}))
 
     return redirect(reverse('module:sequence-item', kwargs={'pk': sequence_item_id}))
+
+
+def create_session(request):
```
idegtiarov: This is odd view function it is not needed.

bridge_adaptivity/bridge_lti/urls.py
```diff
@@ -1,9 +1,10 @@
 from django.conf.urls import url
 
 from bridge_lti.consumer import source_preview
-from bridge_lti.provider import lti_launch
+from bridge_lti.provider import lti_launch, create_session
```
idegtiarov: I think we could propose student to visit index bridge's page for new session creation.

bridge_adaptivity/bridge_lti/urls.py
```diff
@@ -1,9 +1,10 @@
...
 
 urlpatterns = [
     url(r'^launch/?(?P<collection_id>\d*)/?$', lti_launch, name='launch'),
     url(r'^source/$', source_preview, name='source-preview'),
+    url(r'^create_session/$', create_session),
```
idegtiarov: New endpoint is not needed as for me.

bridge_adaptivity/module/mixins.py
```diff
@@ -30,7 +31,12 @@ def dispatch(self, request, *args, **kwargs):
         sequence_id = request.session.get('Lti_sequence')
         if not lti_session:
             log.error('Lti session is not found, Request cannot be processed')
-            return HttpResponseForbidden("Cource content is available only through LTI protocol.")
+            msg = """
```
idegtiarov: To ensure the unit is opened through the LTI initial string must be left and extended with your text. I would only exchange the URL to BRIDGE_HOST. And add possibility with cookies security decreasing to allow all for start working on the adaptive unit and increase it back after unit starts to render correctly.

@kunanit Andrew could you please explain why do we need to visit `https://courses.openedx.vpal.io/event` before new bridge session creation?

bridge_adaptivity/module/mixins.py
```diff
@@ -30,7 +31,12 @@ def dispatch(self, request, *args, **kwargs):
...
+Your browser security settings require an additional step before starting the quiz.
```
idegtiarov: Please improve indentation for the multi-string message.

bridge_adaptivity/bridge_lti/urls.py
kunanit: Index page also works too - My reasoning for making this new view was so that learners didn't get confused by seeing the login screen for the bridge when they went to the index page.

bridge_adaptivity/module/mixins.py
kunanit: It appeared that once the learner has a session by visiting the bridge in a new tab, are able to access the sequence view, and they try to submit a new problem, clicking the "Submit" button (this is in the content source iframe) appears to not do anything. On click, it triggers these errors in the browser console:

screen shot 2017-10-12 at 11 04 57 am

I figured it was the same kind of problem, so visiting the content source in a new tab and reloading the LTI tool solved this. I chose `/event` on the content source so that learners didn't get confused and try to sign up for accounts on the content source openedx instance.

idegtiarov: Oh, got it, thanks. I didn't know that OpenEdx uses the session to check submitted answers.

alexbojko: it should be better to raise exception (in case if it should return HTML response, but with status code for example 400 or 403)

```
from django.core.exceptions import  SuspiciousOperation, PermissionDenied, Http404
from django.http.response import Http404

# to return HttpResponse with status code 400
raise SuspiciousOperation("Bad request")
# you may open django.core.exceptions file and find exceptions inherited from 
# SuspiciousOperation class, they are also OK to convert to response 400.

# to return HttpResponse with status code 403
raise PermissionDenied('You can not view this content')

# to return HttpResponse with status code 404
raise Http404('Page not found')

# to return HttpResponse with status code 500
raise Exception()
```

django will catch exception and handle it properly, so it will return correct response rendered into correct template.

[How django handles errors](https://docs.djangoproject.com/en/2.0/topics/http/urls/#error-handling), and [here](https://docs.djangoproject.com/en/2.0/ref/views/#error-views)

alexbojko:
@kunanit , should we close this PR or should we review and merge it in the future? 
As far as I understand I've already fixed the issue `AD-22`?

PR #70, opened 2017-12-18
Added Collection Groups models, migrations, test for CRU(D?) functionality with this model.

alexbojko:
Added:
* CollectionGroup model
* Engine Model
* Create Read Update views
* added link to Groups in top nav bar
* added basic test cases for views
* templates for list, detail view and update view
* add fixtures for Engine model with mock engine and vpal


bridge_adaptivity/module/models.py
```diff
@@ -110,6 +113,86 @@ def get_absolute_url(self):
...
+    DEFAULT_ENGINE_NAME = 'mock'
+
+    name = models.CharField(default=DEFAULT_ENGINE_NAME, max_length=255)
+    host = models.URLField(blank=True, null=True)
+    token = models.CharField(max_length=255, blank=True, null=True)
+
+    _engines_cache = {}
+
+    class Meta:
+        unique_together = (('host', 'token'),)
```
idegtiarov: wrapped tuple is seemed to be odd

bridge_adaptivity/module/models.py
```diff
@@ -110,6 +113,86 @@ def get_absolute_url(self):
...
+
+    def __str__(self):
+        return "Engine: {}".format(self.name)
+
+    @classmethod
+    def get_default_engine(cls):
+        if not cls.objects.count():
```
idegtiarov: I would use `get_or_create` or you see any restriction to use this shortcut method? BTW I'm not sure that `mock` engine should be default one I think we need to add the flag `default` which could be True only for one engine.

bridge_adaptivity/module/models.py
```diff
@@ -110,6 +113,86 @@ def get_absolute_url(self):
...
+        else:
+            return cls.objects.get(name=cls.DEFAULT_ENGINE_NAME)
+
+    def get_absolute_url(self):
+        return ''
+
+    def make_engine(self):
+        try:
+            engine_module = importlib.import_module('module.engines.engine_{}'.format(self.name.lower()))
+        except ImportError:
```
idegtiarov: With the fixture of pre-set engines try-except block can become useless.

bridge_adaptivity/module/models.py
```diff
@@ -110,6 +113,86 @@ def get_absolute_url(self):
...
+            engine_module = importlib.import_module('module.engines.engine_mock')
+        engine_template = 'Engine{}'
+        driver = getattr(
```
idegtiarov: ditto, with the fixture of the engines `name` should be equal to the driver's name in the engine's module

bridge_adaptivity/module/models.py
```diff
@@ -110,6 +113,86 @@ def get_absolute_url(self):
...
+        )
+        if self.name == 'mock':
+            # mock engine takes no params
+            settings = {}
+        else:
+            settings = {
+                'host': self.host,
+                'token': self.token
+            }
+        key = (self.name, tuple(settings.items()))
```
idegtiarov: Why do we cache the Engine?

bridge_adaptivity/module/models.py
```diff
@@ -110,6 +113,86 @@ def get_absolute_url(self):
...
+    description = models.TextField(blank=True, null=True)
+    atime = models.DateTimeField(auto_now_add=True)
+    owner = models.ForeignKey(BridgeUser)
+    slug = AutoSlugField(
+        null=True,
+        populate_from='name',
+        unique_with=['owner'],
+    )
+
+    collections = models.ManyToManyField('Collection')
```
idegtiarov: quotes are not needed here!

bridge_adaptivity/module/models.py
```diff
@@ -110,6 +113,86 @@ def get_absolute_url(self):
...
+
+    engine = models.ForeignKey('Engine')
```
idegtiarov: ditto

bridge_adaptivity/module/views.py
```diff
@@ -21,11 +21,65 @@
 from module import utils
 from module.forms import ActivityForm
 from module.mixins import CollectionIdToContextMixin, LtiSessionMixin
-from module.models import Activity, Collection, Log, Sequence, SequenceItem
+from module.models import (Activity, Collection, CollectionGroup, Engine, Log, Sequence, SequenceItem)
```
idegtiarov: parentheses are not required in L24

bridge_adaptivity/module/views.py
```diff
@@ -21,11 +21,65 @@
...
+
```
idegtiarov: odd empty line

bridge_adaptivity/module/models.py
alexbojko: I added this fag in other branch

bridge_adaptivity/module/models.py
alexbojko: I don't think so... Admin could configure it not correctly and we should handle it correctly.

bridge_adaptivity/module/models.py
alexbojko: this is fallback in case if misconfigured.

bridge_adaptivity/module/models.py
alexbojko: we discussed it later and decided to cache engines. 

Am I not correctly understand you?

bridge_adaptivity/module/models.py
idegtiarov: ok, we will return to the engine in next PRs

bridge_adaptivity/module/views.py
```diff
@@ -21,11 +21,65 @@
...
+class GroupList(ListView):
+    model = CollectionGroup
+    context_object_name = 'groups'
+    ordering = ['id']
+
+    def get_queryset(self):
+        return self.model.objects.filter(owner=self.request.user)
+
+
+class GroupEditFormMixin(object):
```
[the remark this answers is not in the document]
idegtiarov: I would move all tests into separate dir in the ROOT path like `bridge_adaptivity/tests/module/models_tests.py`

bridge_adaptivity/config/settings/test.py
```diff
@@ -3,16 +3,20 @@
...
 
-DATABASES = {
-    'default': {
-        'ENGINE': 'django.db.backends.postgresql_psycopg2',
-        'NAME': 'travicidb',
-        'USER': 'test',
-        'PASSWORD': 'password',
-        'HOST': 'localhost',
-        'PORT': '5432',
+try:
```
idegtiarov: I would use the same construction as we have in local.py: 
```
try:
    import secure
except ImportError:
    import secure_example as secure
```
And extend default database with the `TEST` block see: https://docs.djangoproject.com/en/2.0/ref/settings/#std:setting-DATABASE-TEST

bridge_adaptivity/module/views.py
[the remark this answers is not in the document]
alexbojko: decided to create dir tests in each application and store tests there.

.travis.yml
```diff
@@ -7,6 +7,18 @@ python:
 install: pip install tox-travis
 script: tox
 
+services:
+  - postgresql
+addons:
+  postgresql: "9.6"
+
+before_script:
+  - psql -c "CREATE DATABASE travicidb;" -U postgres
```
idegtiarov: Here and everywhere s/travicidb/traviscidb/

bridge_adaptivity/bridge_lti/provider.py
```diff
@@ -67,24 +69,68 @@ def instructor_flow(collection_id=None):
...
+    if not collection:
+        log.exception("Collection with provided ID does not exist. Check configured launch url.")
+        raise Http404('Bad launch_url collection ID.')
+
+    collection_group = CollectionGroup.objects.filter(slug=group_slug).first()
+    if collection_group:
+        engine = collection_group.engine or Engine.get_default_engine()
+    else:
+        engine = Engine.get_default_engine()
+    if not collection_group:
```
idegtiarov: L84 is opposite to the L80 and should be handled in the condition for L82

bridge_adaptivity/bridge_lti/tests/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'Boiko'
```
idegtiarov: do we need this line? All authors already mentioned in the AUTHORS file?

bridge_adaptivity/config/settings/secure_example.py
```diff
@@ -20,6 +20,11 @@
         'PASSWORD': 'bridge-adaptivity',
         'HOST': 'postgres',
         'PORT': 5432,
+        'TEST': {
+            'NAME': 'travicidb',
+            'USER': 'bridge',
```
idegtiarov: if USER is not changed for the TEST database you can omit L25

bridge_adaptivity/config/settings/secure_example.py
```diff
@@ -20,6 +20,11 @@
         'PASSWORD': 'bridge-adaptivity',
         'HOST': 'postgres',
         'PORT': 5432,
+        'TEST': {
+            'NAME': 'travicidb',
```
idegtiarov: s/travicidb/traviscidb/

bridge_adaptivity/config/settings/secure_example.py
```diff
@@ -20,6 +20,11 @@
...
+            'USER': 'bridge',
+            'PASSWORD': 'bridge-adaptivity',
```
idegtiarov: ditto password is not changed, but HOST does please set up correct HOST variable.

bridge_adaptivity/config/settings/test.py
```diff
@@ -1,17 +1,15 @@
...
-    'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': 'test.db',
-    }
-}
+try:
+    import secure
+    DATABASES = secure.DATABASES
+except ImportError:
+    import secure_example
```
idegtiarov: please use `as` this helps to avoid code duplication:
```
try:
     import secure
 except ImportError:
     import secure_example as secure
DATABASES = secure.DATABASES
```

bridge_adaptivity/module/urls.py
```diff
@@ -4,14 +4,24 @@
...
-    url(r'^collection/$', CollectionList.as_view(), name='collection-list'),
-    url(r'^collection/add/$', CollectionCreate.as_view(), name='collection-add'),
-    url(r'^collection/(?P<pk>\d+)/change/$', CollectionUpdate.as_view(), name='collection-change'),
-    url(r'^collection/(?P<pk>\d+)/$', CollectionDetail.as_view(), name='collection-detail'),
+    url(r'^group/$', GroupList.as_view(), name='group-list'),
+    url(r'^group/add/?$', GroupCreate.as_view(), name='group-add'),
+    url(r'^group/(?P<pk>\d+)/$', GroupDetail.as_view(), name='group-detail'),
+    url(r'^group/(?P<pk>\d+)/change/$', GroupUpdate.as_view(), name='group-change'),
+
+    url(r'^(?:group/(?P<group_slug>\w+)/)?collection/$', CollectionList.as_view(), name='collection-list'),
```
idegtiarov: L17-23 you use the two same tools please choose one of them `(?:group/(?P<group_slug>\w*)/)` or `(group/(?P<group_slug>\w+)/)?`

bridge_adaptivity/module/views.py
```diff
@@ -20,24 +20,74 @@
...
+
+@method_decorator(login_required, name='dispatch')
+class CollectionList(CollectionMixin, ListView):
     model = Collection
     context_object_name = 'collections'
     ordering = ['id']
 
     def get_queryset(self):
-        return Collection.objects.filter(owner=self.request.user)
+        qs = Collection.objects.filter(owner=self.request.user)
```
idegtiarov: You've already moved this code to the mixin and it is odd here or I missed something.

bridge_adaptivity/bridge_lti/tests/__init__.py
alexbojko: it was added by pycharm, not by me... I will delete it

bridge_adaptivity/module/views.py
alexbojko: fixed. I don't know why git hub did not remove your comment from conversation, but this line is not present anymore in code.

PR #74, opened 2018-01-11
Add multiple engines feature

idegtiarov:
Add feature for dynamically choosing the adaptive engine for the existing collection groups.

The administrator is able to configure adaptive engines which are based on the existing engine drivers.
The instructor is able to add created by the administrator engines to the collection groups.


bridge_adaptivity/module/models.py
```diff
@@ -13,12 +15,33 @@
...
+def _get_engine_driver(engine):
+    driver = None
+    engine_module = importlib.import_module('module.engines.{}'.format(engine))
+    for attr in inspect.getmembers(engine_module):
+        if attr[0].startswith('Engine'):
+            driver = attr[1]
+    return driver
+
+
+ENGINES = _discover_engines()
```
alexbojko: as far as I remember we discussed that user should be able to add new engines without server reloading? isn't it? 

If I correctly understand what's going on here - user will not be able to add new one engine on the fly. 

correct me If I missed something.

bridge_adaptivity/module/models.py
```diff
@@ -116,49 +140,42 @@ def get_absolute_url(self):
...
 
     class Meta:
         unique_together = ('host', 'token')
 
     def __str__(self):
-        return "Engine: {}".format(self.name)
+        return "Engine: {}".format(self.engine_name)
+
+    def save(self, *args, **kwargs):
+        utils.save_model_parameter_true_once(self, 'is_default')
```
alexbojko: little bit not clear function name as for me. 

I understand that you save model with flag is_default=True and then update all other objects in DB with is_default=False, but I understand it only after opening this function and reading code.

In my branch I created such mixing:
```
class ModelWithDefaultInstanceMixin(object):
    IS_DEFAULT_FIELD = 'is_default'

    def save(self, *args, **kwargs):
        if not hasattr(self, self.IS_DEFAULT_FIELD):
            raise ValueError('Model {} has no field {}'.format(self, self.IS_DEFAULT_FIELD))

        if getattr(self, self.IS_DEFAULT_FIELD, None):
            default_qs = self.__class__.objects.filter(**{self.IS_DEFAULT_FIELD: True})
            if default_qs:
                default_qs.update(is_default=False)
        return super(ModelWithDefaultInstanceMixin, self).save(*args, **kwargs)
```

don't you think it's better solution?

alexbojko: And using this mixin you don't need to define save for each model as you do here in this PR.

bridge_adaptivity/module/models.py
```diff
@@ -116,49 +140,42 @@ def get_absolute_url(self):
...
-        # otherwise create Engine, put it in cache and return it
-        engine_module = importlib.import_module('module.engines.engine_{}'.format(self.name.lower()))
-        engine_template = 'Engine{}'
-        driver = getattr(
-            engine_module, engine_template.format(self.name.upper()),  # try uppercase
-            getattr(
-                engine_module,
-                engine_template.format(self.name.capitalize()),  # if not found - try capitalized
-            )
+        return (
```
alexbojko: I would avoid parentheses and reformat this code in a such way:
```
return Engine.objects.filter(
    is_default=True
).first() or Engine.objects.get_or_create(
    engine=cls.DEFAULT_ENGINE, engine_name='Mock', is_default=True
)[0]
```

but probably we should discuss it...

bridge_adaptivity/module/models.py
```diff
@@ -116,49 +140,42 @@ def get_absolute_url(self):
...
+            Engine.objects.get_or_create(engine=cls.DEFAULT_ENGINE, engine_name='Mock', is_default=True)[0]
         )
-        Engine._engines_cache[key] = driver(**settings)
-        return Engine._engines_cache[key]
+
+    @property
+    def engine_driver(self):
+        if not self.DRIVER:
+            driver = _get_engine_driver(self.engine)
+            if self.engine.endswith('mock'):
```
alexbojko: I understand that all modules in python should be named in lower case, but.... 
I would use  ```self.engine.lower()``` just to be sure 
now we have only 2 engines, and both of them written by us or we reviewed this code. from other side if someone would add new engine he could name it win camelcase notation and it will broke this function. 
for example ```engine_VPAL_Mock.py```.

what do you think?

bridge_adaptivity/module/models.py
```diff
@@ -235,30 +252,34 @@ def save(self, *args, **kwargs):
         """Extension which sends notification to the Adaptive engine that Activity is created/updated."""
         initial_id = self.id
         if initial_id:
-            if not ENGINE.update_activity(self):
-                raise ValidationError
+            for engine in self.collection.collection_groups.all():
+                if not engine.update_activity(self):
+                    raise ValidationError
```
alexbojko: you should raise instance if exception, not a class.
raising exception class were in old python versions, and now you should initialize it.

according to [this](https://stackoverflow.com/questions/16706956/is-there-a-difference-between-raise-exception-and-raise-exception-without) there's no difference between these methods, but in my opinion it would be more clear to raise exception some text message which will clarify why did you raise it.

let's discuss it?

bridge_adaptivity/module/models.py
```diff
@@ -235,30 +252,34 @@ def save(self, *args, **kwargs):
...
             )
         super(Activity, self).save(*args, **kwargs)
         if not initial_id:
-            if not ENGINE.add_activity(self):
-                super(Activity, self).delete(*args, **kwargs)
-                raise ValidationError
+            for engine in self.collection.collection_groups.all():
+                if not engine.add_activity(self):
+                    super(Activity, self).delete(*args, **kwargs)
+                    raise ValidationError
```
alexbojko: ditto

bridge_adaptivity/module/models.py
```diff
@@ -235,30 +252,34 @@ def save(self, *args, **kwargs):
...
-        if not ENGINE.delete_activity(self):
-            raise ValidationError
-        Log.objects.create(
-            log_type=Log.ADMIN, action=Log.ACTIVITY_DELETED,
-            data=self.get_research_data()
-        )
+        for engine in self.collection.collection_groups.all():
+            if not engine.delete_activity(self):
+                # Note(idegtiarov) Add handler for the case of partial engines synchronization
+                raise ValidationError
```
alexbojko: ditto

bridge_adaptivity/module/tests/test_models.py
```diff
@@ -0,0 +1,56 @@
...
+        engine_default_set = Engine.objects.filter(is_default=True)
+        self.assertFalse(engine_default_set)
+        default_engine = Engine.get_default_engine()
+        engine_default_new_set = Engine.objects.filter(is_default=True)
+        self.assertEquals(engine_default_new_set.count(), 1)
+        self.assertEquals(engine_default_set.first(), default_engine)
+
+    def test_engine_driver_property(self):
+        """Test Engine model's property engine_driver."""
+        host, token = ('fake_host', 'fake_token')
```
alexbojko: would avoid to use parentheses

tox.ini
```diff
@@ -30,4 +30,4 @@ max-line-length = 120
 import-order-style = google
 # Ignoring requirements of the docstrings in public packages, classes, functions, etc.
 ignore = D100, D101, D102, D103, D104, D105, D106, D107
-application-import-names = module, api, bridge_lti
```
alexbojko: I see why did you add this line, but I prefer to use the whole path to module e.g. ```from module.models import Collection``` or ```from module.views import SomeView```.

Also, I had some troubles in tests when I used local imports... 

but it's only my opinion and we could discuss it...

bridge_adaptivity/common/utils.py
```diff
@@ -0,0 +1,7 @@
+# Common for modules functions are gathering in utils module
+
+
+def save_model_parameter_true_once(instance, parameter):
+    if getattr(instance, parameter, None):
+        filter_kwargs, update_kwargs = ({parameter: True}, {parameter: False})
```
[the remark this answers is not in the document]
idegtiarov: Parensethes are used for the clarity and I agree it is not required will remove them.

bridge_adaptivity/module/admin.py
```diff
@@ -51,7 +51,7 @@ class CollectionGroupAdmin(admin.ModelAdmin):
 
 @admin.register(Engine)
 class EngineAdmin(admin.ModelAdmin):
-    readonly_fields = ['name', 'host', 'token']
```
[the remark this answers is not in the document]
idegtiarov: Not it should be editable. Discussed locally.

bridge_adaptivity/module/models.py
idegtiarov: New Engine could be created dynamically `_discover_engines()` returns engines libs with the driver API implementation.

bridge_adaptivity/module/models.py
idegtiarov: Mixin works for me. We will improve this code while merging with your branch.

bridge_adaptivity/module/models.py
idegtiarov: I would leave as it is now if you don't mind

bridge_adaptivity/module/models.py
idegtiarov: It is coming from the driver's lib namespace in engines package and currently covers existent state I will leave a comment to pay attention to this part of the code.

bridge_adaptivity/module/models.py
idegtiarov: It would be removed in the coming PR with the engine synchronization.

bridge_adaptivity/module/tests/test_models.py
idegtiarov: done

tox.ini
idegtiarov: It is not like you described, here modules are added for the config to let tox recognize internal project's modules.

PR #77, opened 2018-01-23
Multiple Engines: Sync between Engine and Collection

idegtiarov:
Add sync procedure for dynamical engine changing in the collection group as async Celery task.


alexbojko:
could you please update this branch with latest changes from master branch

PR #78, opened 2018-01-24
Multiple Engines add Sync API endpoint

idegtiarov:
Add new endpoint for collection synchronization with engines.
Remove unused CRUD endpoints.
Add tests.


bridge_adaptivity/config/celery.py
```diff
@@ -0,0 +1,19 @@
...
+# - namespace='CELERY' means all celery-related configuration keys
+#   should have a `CELERY_` prefix.
+celery_app.config_from_object('django.conf:settings', namespace='CELERY')
+
+# Load task modules from all registered Django celery_app configs.
+celery_app.autodiscover_tasks()
+
+
+@celery_app.task(bind=True)
+def debug_task(self):
```
alexbojko: Do we really need this debug task?

bridge_adaptivity/config/settings/base.py
```diff
@@ -128,6 +128,17 @@
     },
 }
 
+# Celery settings
+
+# Timespan for running sync task in seconds
+CELERY_DELAY_SYNC_TASK = 300  # default value is equal to 5 minutes
```
alexbojko: I would be more clear to write like this 
```CELERY_DELAY_SYNC_TASK = 5 * 60  # default value is equal to 5 minutes```

File: bridge_adaptivity/module/engines/engine_vpal.py [hunk 3]
```diff
@@ -143,4 +113,21 @@ def submit_activity_answer(self, sequence_item):
         submit_url = urlparse.urljoin(self.base_url, 'score')
         payload = self.fulfill_payload(instance_to_parse=sequence_item)
         submit_activity_score = requests.post(submit_url, json=payload, headers=self.headers)
-        return self.check_engine_response(submit_activity_score.status_code, 'graded', sequence_item.activity.name)
+        return self.check_engine_response(
+            submit_activity_score.status_code, action='graded', obj='sequence item', name=sequence_item.activity.name
```
alexbojko: trailing comma?

bridge_adaptivity/module/engines/engine_vpal.py
```diff
@@ -143,4 +113,21 @@ def submit_activity_answer(self, sequence_item):
...
+
+        :param collection: Collection instance for synchronization
+        """
+        sync_url = urlparse.urljoin(self.base_url, 'sync/collection/{}'.format(collection.id))
+        payload = {"collection": collection.id, "activities": []}
+        for activity in collection.activities.all():
+            payload["activities"].append(self.fulfill_payload(instance_to_parse=activity))
+        sync_collection = requests.post(sync_url, json=payload, headers=self.headers)
+        return self.check_engine_response(
+            sync_collection.status_code, action='synchronized', obj='collection', name=collection.name
```
alexbojko: comma?

bridge_adaptivity/module/migrations/0007_collection_updated_at.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
alexbojko: In my opinion we should create abstract model with fields ```created_at``` and ```updated_at``` and inherit mostly all models from this model. 

It will give us ability to know when objects were created and updated.

bridge_adaptivity/module/tasks.py
```diff
@@ -0,0 +1,14 @@
+from __future__ import absolute_import, unicode_literals
+
+from celery.task import task
+
+
+@task()
+def sync_collection_engines(collection_id=None, created_at=None):
+    from module.models import Collection
```
alexbojko: why don't you use import on module level?

bridge_adaptivity/module/tests/test_tasks.py
```diff
@@ -0,0 +1,35 @@
+from django.conf import settings
+from django.test import TestCase
+from mock import patch
+
+from bridge_lti.models import BridgeUser
+from module.models import Activity, Collection, CollectionGroup, Engine
+from module.tasks import sync_collection_engines
+
+
+class TestTask(TestCase):
```
alexbojko: you could inherit this class from ```module/tests/test_views.py:BridgeTestCase```.
in this class method setup already implemented all lines which you wrote here.

pls take a look into this class. 

btw maybe we should place it not in ```module/tests``` but somewhere in general folder and re-name it?

bridge_adaptivity/module/tests/test_views.py
```diff
@@ -15,7 +16,8 @@ def add_prefix(self, prefix='', data={}):
         """Add prefix to form data dict, which will be send as POST or GET to view."""
         return {"{}-{}".format(prefix, k): v for k, v in data.items()}
 
-    def setUp(self):
+    @patch('module.tasks.sync_collection_engines.apply_async')
```
alexbojko: I can't see where do you use this mock?

bridge_adaptivity/config/celery.py
idegtiarov: It is not needed. Removed.

bridge_adaptivity/config/settings/base.py
idegtiarov: Thanks, that is a nice catch. Improved.

bridge_adaptivity/module/engines/engine_vpal.py
idegtiarov: Not in this case. Within any changes in this line, it will be added to the commit diff view anyway.

bridge_adaptivity/module/engines/engine_vpal.py
idegtiarov: ditto

bridge_adaptivity/module/migrations/0007_collection_updated_at.py
idegtiarov: We will think about it to figure out which models require timestamp fields. And if find that it would be helpful will add them in one of the following PR.

bridge_adaptivity/module/tasks.py
idegtiarov: Because of cycling import, unfortunately, we use this task in the module.models file.

bridge_adaptivity/module/tests/test_tasks.py
idegtiarov: The inheritance does not make sense, in setUp collection creation need to be avoided and added only to the test.

PR #79, opened 2018-01-24
Add source_launch_url as an identifier for the activity in the engine.

idegtiarov:
Identifier for the Activity is changed to the source_launch_url
Tests are added.


alexbojko:
pls merge conflicts

bridge_adaptivity/module/tests/test_utils.py
```diff
@@ -0,0 +1,46 @@
+from django.test import TestCase
+from mock import patch
+
+from bridge_lti.models import BridgeUser, LtiProvider, LtiUser, OutcomeService
+from module.models import Activity, Collection, Engine, GradingPolicy, Sequence
+from module.utils import choose_activity
+
+
+class TestUtilities(TestCase):
+    fixtures = ['gradingpolicy', 'engine']
```
alexbojko: I think it would be better to write the whole name of fixture with file extension. 
It give us an ability to find this fixtures much faster.

```
fixtures = ['gradingpolicy.json', 'engine.json']
```

alexbojko: LGTM, fix about fixtures naming is up to you.

PR #85, opened 2018-02-01
Fix the bug #AD-67

idegtiarov:
Fix redirection on the update/create views and add back button.


idegtiarov:
Good question. Will update PR.

bridge_adaptivity/module/mixins.py
```diff
@@ -97,3 +97,12 @@ class CollectionMixin(object):
     def get_queryset(self):
         qs = Collection.objects.filter(owner=self.request.user)
         return qs
+
+
+class BackURLMixin(object):
+    def get_context_data(self, **kwargs):
```
alexbojko: probably it should be better to define field name which should be used for back button.
in some cases it could be the same as used in the form and in this case we will change it in the view.

```
class BackURLMixin(object):
    back_url_field_name = 'back_url'

    def get_context_data(self, **kwargs):
        context = super(BackURLMixin, self).get_context_data(**kwargs)
        back_url = self.request.GET.get(self.back_url_field_name)
        if back_url:
            context['back_url'] = back_url
        return context
```

bridge_adaptivity/module/templates/module/collection_form.html
```diff
@@ -16,7 +16,7 @@ <h2>Edit collection: {{collection.name }}</h2>
   {% bootstrap_form form %}
   <button class="btn btn-primary" type="submit">
     {% if collection.id %}Update{% else %}Create{% endif %}</button>
-  <a href="{% if collection.id %}{% url 'module:collection-detail' collection.id %}{% else %}{% url 'module:collection-list' %}{% endif %}" class="btn btn-default" role="button"
+  <a href="{% if back_url or not collection.id %}{% url 'module:collection-list' %}{% else %}{% url 'module:collection-detail' collection.id %}{% endif %}" class="btn btn-default" role="button"
```
alexbojko: probably it should be better to move this logic to View class ?

bridge_adaptivity/module/templates/module/collectiongroup_form.html
```diff
@@ -19,7 +19,7 @@ <h2>New Collection Group configuration</h2>
   </div>
   <button class="btn btn-primary" type="submit">
     {% if group.id %}Update{% else %}Create{% endif %}</button>
-  <a href="{% if group.id %}{{ group.get_absolute_url }}{% else %}{% url 'module:group-list' %}{% endif %}" class="btn btn-default" role="button"
+  <a href="{% if back_url or not group.id %}{% url 'module:group-list' %}{% else %}{{ group.get_absolute_url }}{% endif %}" class="btn btn-default" role="button"
```
alexbojko: should be better to move logic to view?

bridge_adaptivity/module/templates/module/collection_form.html
idegtiarov: Agree.

bridge_adaptivity/module/templates/module/collectiongroup_form.html
idegtiarov: ditto

bridge_adaptivity/module/mixins.py
idegtiarov: It is not clear enough what is the pros in proposed change?

PR #86, opened 2018-02-02
Non problem activities

alexbojko:
### [SOLUTION]
* changes in mock engine for correct work with `non problem` activities;
* add field `Activity.stype` to store activity type (video, html or problem);
* add field `SequenceItem.is_problem` to be able to filter sequence items by this field and determine whether sequence item is problem or not;
* add `module.models.PROBLEM_ACTIVITY_TYPES` which is a list of types of `blocks` (edx), which are require grading;
* updated `module.js` JS to pass type to `Activity.stype` field when create activity 

#### Working on test for this functionality


bridge_adaptivity/module/forms.py
```diff
@@ -14,6 +14,9 @@ class ActivityForm(ModelForm):
     class Meta:
         model = Activity
         exclude = ['collection', 'lti_consumer']
+        widgets = {
+            'type': forms.HiddenInput(),
```
idegtiarov: We've decided to name this parameter as `stype`

bridge_adaptivity/module/models.py
```diff
@@ -52,11 +52,17 @@ def _load_cls_from_applicable_module(module_path, mod_name, class_startswith=Non
 }
 GRADING_POLICY_CHOICES = ((k, v.public_name) for k, v in GRADING_POLICY_NAME_TO_CLS.items())
 
+PROBLEM_ACTIVITY_TYPES = [
```
idegtiarov: I would use tuple here instead of a list.

bridge_adaptivity/module/models.py
```diff
@@ -52,11 +52,17 @@ def _load_cls_from_applicable_module(module_path, mod_name, class_startswith=Non
...
+    'problem',
+]
+
 
 @python_2_unicode_compatible
 class Sequence(models.Model):
     """Represents User's problem solving track."""
 
+    PROBLEM_ACTIVITY_TYPES = PROBLEM_ACTIVITY_TYPES
```
idegtiarov: Why line 64 was added?

bridge_adaptivity/module/tests/test_models.py
```diff
@@ -115,3 +119,68 @@ def test_policy_cls_property(self):
...
+          )
+    def test_create_sequence_item_for_activity(self, stype, is_problem):
+        activity = Activity.objects.create(
+            name='test', collection=self.collection1, tags='test', atype='G', stype=stype
+        )
+        sequence_item = SequenceItem.objects.create(
+            sequence=self.sequence,
+            activity=activity,
+        )
+        self.assertTrue(sequence_item.is_problem == activity.is_problem)
```
idegtiarov: What do we test in L185 if it is the same by the definition?

bridge_adaptivity/module/models.py
```diff
@@ -281,6 +289,15 @@ class Activity(OrderedModel):
     source_context_id = fields.CharField(max_length=255, blank=True, null=True)
     # NOTE(wowkalucky): extra field 'order' is available (inherited from OrderedModel)
 
+    # `stype` - means source_type or string_type.
+    stype = models.CharField(
+        "Type of activity", help_text="(problem, video, html, etc.)", max_length=25, blank=True, null=True
```
idegtiarov: s/Type of activity/Type of the activity/

bridge_adaptivity/module/forms.py
alexbojko: in model it's already has name stype. forgot to change it here...

bridge_adaptivity/module/models.py
alexbojko: to be able to get problem activity types from sequence instance. I use it in mock engine

bridge_adaptivity/module/models.py
alexbojko: should I move this tuple to config/settings/base.py or somewhere in settings ? I think it should be there....

bridge_adaptivity/module/tests/test_models.py
alexbojko: here we test that sequence has the same value for property is_problem as well as activity.is_problem. 

Also, it checks that logic of save method of Sequence object is working correctly.

bridge_adaptivity/module/tests/test_mock_engine.py
```diff
@@ -0,0 +1,104 @@
...
+            _activity['source_launch_url'] = activity['source_launch_url'].format(i)
+            activities.append(Activity.objects.create(
+                name='test_{}'.format(i), collection=self.collection1, tags='test', **_activity
+            ))
+
+        source_launch_urls = Activity.objects.filter(
+            id__in=[i.id for i in activities]
+        ).values_list('source_launch_url', flat=True)
+
+        created_sequence_items = []
```
idegtiarov: `created_sequence_items` is declared but not used.

bridge_adaptivity/module/tests/test_mock_engine.py
```diff
@@ -0,0 +1,104 @@
...
+        self.test_cg.collections.add(self.collection1)
+        self.sequence = Sequence.objects.create(
+            lti_user=self.lti_user, collection=self.collection1,
+            engine=self.engine, grading_policy=self.trials_count,
+        )
+
+    def test_check(self):
+        self.assertTrue(True)
+
+    @patch('module.tasks.sync_collection_engines.apply_async')
```
idegtiarov: I would set @patch decorator to be the first decorator under the test

bridge_adaptivity/module/tests/test_models.py
idegtiarov: got it. Thanks

bridge_adaptivity/module/tests/test_mock_engine.py
```diff
@@ -0,0 +1,104 @@
...
+        self.consumer = LtiProvider.objects.create(
+            consumer_name='name',
+            consumer_key='key',
+            consumer_secret='secret',
+        )
+        self.lti_user = LtiUser.objects.create(
+            user_id='some_user', course_id='some_course', email=self.user.email,
+            lti_consumer=self.consumer, bridge_user=self.user
+        )
+        self.client.login(username='test', password='test')
```
idegtiarov: why do we need client authorization? I cannot find where it is used in the tests.

bridge_adaptivity/module/tests/test_models.py
```diff
@@ -115,3 +119,70 @@ def test_policy_cls_property(self):
...
+        self.consumer = LtiProvider.objects.create(
+            consumer_name='name',
+            consumer_key='key',
+            consumer_secret='secret',
+        )
+        self.lti_user = LtiUser.objects.create(
+            user_id='some_user', course_id='some_course', email=self.user.email,
+            lti_consumer=self.consumer, bridge_user=self.user
+        )
+        self.client.login(username='test', password='test')
```
idegtiarov: ditto

bridge_adaptivity/module/tests/test_models.py
```diff
@@ -115,3 +119,70 @@ def test_policy_cls_property(self):
...
+            engine=self.engine,
+            grading_policy=self.points_earned
+        )
+        self.test_cg.collections.add(self.collection1)
+        self.sequence = Sequence.objects.create(
+            lti_user=self.lti_user, collection=self.collection1,
+            engine=self.engine, grading_policy=self.trials_count,
+        )
+
+    @patch('module.tasks.sync_collection_engines.apply_async')
```
idegtiarov: ditto

bridge_adaptivity/module/tests/test_models.py
```diff
@@ -115,3 +119,70 @@ def test_policy_cls_property(self):
...
+    @unpack
+    @data({'stype': 'video', 'is_problem': False},
+          {'stype': 'html', 'is_problem': False},
+          {'stype': 'problem', 'is_problem': True},
+          )
+    def test_is_problem_property(self, mock_apply_async, stype, is_problem):
+        activity = Activity(name='test', collection=self.collection1, tags='test', atype='G', stype=stype)
+        self.assertEqual(activity.is_problem, is_problem)
+
+    @patch('module.tasks.sync_collection_engines.apply_async')
```
idegtiarov: ditto

PR #93, opened 2018-02-07
Fix error 500 (AD-72)

alexbojko:
### [PROBLEM]
is described [here](https://vpalresearch.atlassian.net/browse/AD-72)

### [SOLUTION]
* Remove fields grading_policy and engine form Sequence model, replace them with field group
* Update unique_together, `unique_together = (('lti_user', 'collection', 'group'), ('lis_result_sourcedid', 'outcome_service'))` - add `group` to unique_together with `lti_user` and `collection`, so now it will get sequence, or will b able to correctly create new one, without error.
* add data migration, which collect all Sequqnces and for each of them tries to find his group based on collection, policy and engine. If found more than 1 group it will show warning message with the link to admin UI, where users will be able to update group after migration process finish. (NOTE that it will not cancel migration)


bridge_adaptivity/module/models.py
```diff
@@ -63,14 +63,17 @@ class Sequence(models.Model):
 
     lti_user = models.ForeignKey(LtiUser)
     collection = models.ForeignKey('Collection')
-    engine = models.ForeignKey('Engine', blank=True, null=True)
```
idegtiarov: L109 needs to be updated either

bridge_adaptivity/module/models.py
```diff
@@ -63,14 +63,17 @@ class Sequence(models.Model):
...
-    grading_policy = models.ForeignKey('GradingPolicy', blank=True, null=True)
+    group = models.ForeignKey('CollectionGroup', null=True)
+
+    # engine = models.ForeignKey('Engine', blank=True, null=True)
```
idegtiarov: odd comments, please remove

PR #94, opened 2018-02-12
Add tests for send_grade method in all policies

alexbojko:
### [PROBLEM]
Described in  ticket AD-90

### [SOLUTION]
Added tests for Policy.send_grade function.

#### [NOTE]
As I found here http://www.voidspace.org.uk/python/mock/patch.html#where-to-patch we should mock no in place where variable or function is defined but in place where it is used.

### DIFF
--- a/bridge_adaptivity/module/tests/test_grading_policy.py
+++ b/bridge_adaptivity/module/tests/test_grading_policy.py
@@ -1,9 +1,14 @@
 from ddt import data, ddt, unpack
 from django.test import TestCase
+from django.test.client import RequestFactory
 import mock
 import pytest
 
-from module.models import GRADING_POLICY_NAME_TO_CLS, GradingPolicy, Sequence
+from bridge_lti.models import LtiProvider, OutcomeService
+from module.models import (
+    Activity, BridgeUser, Collection, CollectionGroup, Engine, GRADING_POLICY_NAME_TO_CLS, GradingPolicy, LtiUser,
+    Sequence
+)
 from module.policies.policy_full_credit import FullCreditOnCompleteGradingPolicy
 from module.policies.policy_points_earned import PointsEarnedGradingPolicy
 from module.policies.policy_trials_count import TrialsCountGradingPolicy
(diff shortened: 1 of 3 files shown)


bridge_adaptivity/module/tests/test_grading_policy.py
```diff
@@ -105,3 +110,84 @@ def test_policy_math(
...
+        )
+        self.test_cg.collections.add(self.collection)
+        self.sequence = Sequence.objects.create(
+            lti_user=self.lti_user,
+            collection=self.collection,
+            group=self.test_cg,
+            outcome_service=self.outcome_service
+        )
+
+    @mock.patch('module.policies.policy_points_earned.update_lms_grades')
```
idegtiarov: I would change the import of the `update_lms_grades ` function in the policies from:
`from bridge_lti.outcomes import update_lms_grades` to:
`from bridge_lti import outcomes` and use the function as `outcomes.update_lms_grades` so we could patch required function directly in the parent module and continue using `ddt` library to unpack tests data.

bridge_adaptivity/module/tests/test_grading_policy.py
```diff
@@ -105,3 +110,84 @@ def test_policy_math(
...
+        )
+        default_kw = {
+            'sequence': self.sequence,
+            'user_id': self.lti_user.user_id,
+        }
+        policy.send_grade()
+        mock_update_lms_grades.assert_called_with(*(request,), **default_kw)
+
+    @mock.patch('module.policies.base.update_lms_grades')
+    def test_send_grade_method_policy_trials_count(self, mock_update_lms_grades):
```
idegtiarov: please rewrite test name to cover all tests cases.

bridge_adaptivity/module/tests/test_grading_policy.py
alexbojko: discussed with @idegtiarov and decided to not change this part of code and follow django style of imports.

PR #97, opened 2018-02-13
(fix/AD-96): Add course, create, update, delete, detail views, tests for these views.

alexbojko:
### [PROBLEM]
Is described in ticket [AD-96](https://vpalresearch.atlassian.net/browse/AD-96)

### [SOLUTION]
* add model Course
* update CollectionGroup model, add FK to Course
* add view for CRUD Course
* add tests for these views


bridge_adaptivity/module/mixins/models.py
```diff
@@ -0,0 +1,33 @@
+# coding: utf-8
+from uuid import uuid4
+
+
+class WithUniqueSlugMixin(object):
```
idegtiarov: It seems that this mixing can be omitted with the usage of `UUIDField` https://docs.djangoproject.com/en/2.0/ref/models/fields/#uuidfield

bridge_adaptivity/module/models.py
```diff
@@ -101,6 +101,21 @@ def save(self, *args, **kwargs):
         super(SequenceItem, self).save(*args, **kwargs)
 
 
+@python_2_unicode_compatible
+class Course(WithUniqueSlugMixin, models.Model):
+    name = models.CharField(max_length=255)
+    description = models.TextField(blank=True, null=True)
+
+    slug = models.SlugField(max_length=36)
```
idegtiarov: Lets use UUIDField instead of SlugField, and indexing it.

bridge_adaptivity/module/models.py
```diff
@@ -219,18 +234,17 @@ def engine_driver(self):
...
     owner = models.ForeignKey(BridgeUser)
-    slug = AutoSlugField(
-        null=True,
-        populate_from='name',
-        unique_with=('atime',),
-    )
+    slug = models.SlugField(null=True)
+    course = models.ForeignKey(Course, related_name='course_groups', null=True)
+
+    course = models.ForeignKey(Course, related_name='course_groups', null=True)
```
idegtiarov: duplicate L245

bridge_adaptivity/module/tests/test_models.py
```diff
@@ -203,3 +205,70 @@ def test_update_sequence_item_with_grade(self, mock_apply_async):
         sequence_item.save()
         # save() method is overloaded in the models, and we test that it works correctly
         self.assertEqual(sequence_item.score, 0.5)
+
+
+class TestWithUniqueSlugMixin(TestCase):
```
idegtiarov: within UUID field all this tests won't be needed any more

bridge_adaptivity/module/migrations/0010_auto_20180216_1150.py
```diff
@@ -0,0 +1,87 @@
...
+                ('description', models.TextField(blank=True, null=True)),
+                ('slug', models.UUIDField(default=uuid.uuid4, editable=False, serialize=False, unique=True)),
+                ('owner', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to=settings.AUTH_USER_MODEL)),
+            ],
+        ),
+        migrations.RunPython(forward_function, lambda *a, **kw: None),
+        migrations.AlterField(
+            model_name='collectiongroup',
+            name='slug',
+            field=models.UUIDField(default=uuid.uuid4, editable=False, unique=True),
```
kunanit: I had some trouble running this migration on bridge instances with pre-existing collection groups (that had text field as the slug). It threw an error like: `django.db.utils.DataError: invalid input syntax for uuid: "my-new-group"`. I think it would be good to include a data migration for this, to generate a new field value with the uuid format for those collection group instances that already had an (invalid) text value for the field.

PR #99, opened 2018-02-14
fix/AD-89: User should be able to create empty Group

alexbojko:
### [PROBLEM] 
Is described in ticket [AD-89](https://vpalresearch.atlassian.net/browse/AD-89)

### [SOLUTION]
Add `blank=True` to M2M field `collections` of there `CollectionGroup` model.


bridge_adaptivity/module/tests/test_models.py
```diff
@@ -203,3 +203,20 @@ def test_update_sequence_item_with_grade(self, mock_apply_async):
...
+        user = BridgeUser.objects.create_user(
+            username='test',
+            password='test',
+            email='test@me.com'
+        )
+        points_earned = GradingPolicy.objects.get(name='points_earned')
+        engine = Engine.objects.create(engine='engine_mock')
+        CollectionGroup.objects.create(
+            name='some name', engine=engine, grading_policy=points_earned, owner=user
+        )
```
idegtiarov: It would be nice if test contains at least one assert case otherwise it is not clear what is tested.

PR #103, opened 2018-02-19
Add possibility to sync collection immediately.

idegtiarov:


bridge_adaptivity/module/templates/module/collection_detail.html
```diff
@@ -4,11 +4,18 @@
 {% block title %}
   <span class="label label-default">ID:{{ collection.id }}</span> Collection: {{ collection.name }}
   <h3>
+    {% if sync_available %}
+    <a class="pull-right require-submission"
+       href="{% url 'module:collection-sync' collection.id %}?back_url={{ back_url }}"
```
alexbojko: good good :) my jquery plugin is in use! thats fine!

bridge_adaptivity/module/urls.py
```diff
@@ -57,5 +57,8 @@
     url(r'^$', RedirectView.as_view(url=reverse_lazy('module:collection-list'))),
 
     # Source outcome service endpoint
-    url(r'^callback_grade/$', callback_sequence_item_grade, name='sequence-item-grade')
+    url(r'^callback_grade/$', callback_sequence_item_grade, name='sequence-item-grade'),
```
alexbojko: add `?` in the end of URL to make it work well without trailing slash

` url(r'^callback_grade/?$', callback_sequence_item_grade, name='sequence-item-grade'),`

bridge_adaptivity/module/urls.py
```diff
@@ -57,5 +57,8 @@
...
+
+    # Sync collection with relative engines
+    url(r'^collection/(?P<pk>\d+)/sync/$', sync_collection, name='collection-sync'),
```
alexbojko: ditto

bridge_adaptivity/module/views.py
```diff
@@ -406,3 +406,17 @@ def callback_sequence_item_grade(request):
...
+    Synchronize collection immediately.
+    """
+    back_url = request.GET.get('back_url')
+    collection = Collection.objects.get(pk=pk)
+    collection.save()
+    log.debug("Immediate sync task is created, time: {}".format(collection.updated_at))
+    tasks.sync_collection_engines.delay(
+        collection_id=pk, created_at=collection.updated_at
+    )
+    return redirect(reverse('module:collection-detail', kwargs={'pk': pk}) + '?back_url={}'.format(back_url))
```
alexbojko:
```
In [6]: 'back_url={}'.format(None)
Out[6]: 'back_url=None'
```
if back_url is None it will return not correct url. Am I miss something?

bridge_adaptivity/module/views.py
```diff
@@ -406,3 +406,17 @@ def callback_sequence_item_grade(request):
...
 
     return HttpResponse(xml, content_type="application/xml")
+
+
+def sync_collection(request, pk):
+    """
+    Synchronize collection immediately.
+    """
+    back_url = request.GET.get('back_url')
+    collection = Collection.objects.get(pk=pk)
```
alexbojko: probably it should be better to use  `get_object_or_404` here.

bridge_adaptivity/module/tests/test_views.py
```diff
@@ -377,3 +377,22 @@ def test_update_not_mine_course(self):
...
+
+    @patch('module.tasks.sync_collection_engines.delay')
+    @patch('module.tasks.sync_collection_engines.apply_async')
+    @patch('module.views.get_available_courses')
+    def test_immediate_synchronization(
+        self, mock_get_available_courses, mock_apply_async, mock_delay
+    ):
+        col_id = self.collection1.id
+        expected_url = reverse('module:collection-detail', kwargs={'pk': col_id}) + '?back_url=None'
+        url = reverse('module:collection-sync', kwargs={'pk': col_id})
```
alexbojko: what about negative test case? 

what we will have if someone pass not correct pk?
I would add it just to be sure that it does not fall in 500.

bridge_adaptivity/module/urls.py
idegtiarov: Its internal endpoint which won't be manipulated manually.

bridge_adaptivity/module/urls.py
idegtiarov: ditto

bridge_adaptivity/module/urls.py
alexbojko: ok

bridge_adaptivity/module/views.py
idegtiarov: if back url is None back button won't be added

bridge_adaptivity/module/views.py
idegtiarov: agree, nice catch!

idegtiarov: done

bridge_adaptivity/module/tests/test_views.py
idegtiarov: done

bridge_adaptivity/module/views.py
idegtiarov: After checking back_url is always added in the template.

PR #104, opened 2018-02-20
(fix/engine_grade_policy): Add grading policy that ask grade from engine

alexbojko:
### [PROBLEM]
Is described in ticket [AD-79](https://vpalresearch.atlassian.net/browse/AD-79)

### [SOLUTION]
* add new policy
* when user choose this policy server should check what engine will be used with it
* if engine is not correct (not listed in Policy.require['engine'] it raise form error


bridge_adaptivity/module/engines/engine_vpal.py
```diff
@@ -119,3 +119,10 @@ def sync_collection_activities(self, collection):
         return self.check_engine_response(
             sync_collection.status_code, action='synchronized', obj='collection', name=collection.name, status=201
         )
+
+    def get_grade(self, sequence):
+        url = urlparse.urljoin(self.base_url, 'collection/{collection_id}/grade'.format(
+            collection_id=sequence.collection.id)
+        )
+        response = requests.post(url, json={'learner_id': sequence.owner.id}, headers=self.headers)
```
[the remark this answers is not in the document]
alexbojko: added. check next line

bridge_adaptivity/module/engines/engine_vpal.py
```diff
@@ -121,3 +121,14 @@ def sync_collection_activities(self, collection):
         return self.check_engine_response(
             sync_collection.status_code, action='synchronized', obj='collection', name=collection.name, status=201
         )
+
+    def get_grade(self, sequence):
```
idegtiarov: I would add the description into doc strings to this method.

bridge_adaptivity/module/engines/engine_vpal.py
```diff
@@ -121,3 +121,14 @@ def sync_collection_activities(self, collection):
...
+        url = urlparse.urljoin(self.base_url, 'collection/{collection_id}/grade'.format(
+            collection_id=sequence.collection.id)
+        )
+        response = requests.post(url, json={'learner_id': sequence.owner.id}, headers=self.headers)
+        if self.check_engine_response(response.status_code, status=200):
```
idegtiarov: status=200 is the default parameter and it could be avoided instead I would add `obj` and `action` params to make logs msg become more informative.

bridge_adaptivity/module/engines/engine_vpal.py
```diff
@@ -121,3 +121,14 @@ def sync_collection_activities(self, collection):
...
+            grade = response.json().get('grade')
+            if 0 < grade < 1:
```
idegtiarov: The grade is `0<=grade<=1`

bridge_adaptivity/module/policies/base.py
```diff
@@ -26,6 +26,9 @@ class BaseGradingPolicy(object):
     __metaclass__ = ABCMeta
 
     public_name = 'Grading Policy'
+    require = {
```
idegtiarov: In the base class, an empty dict seems to be fair enough for `require` param

bridge_adaptivity/module/policies/policy_engine_grade.py
```diff
@@ -0,0 +1,42 @@
...
+    @classmethod
+    def get_form_class(cls):
+        from module.forms import BaseGradingPolicyForm
+        return BaseGradingPolicyForm
+
+    def send_grade(self):
+        """Send grade to LMS system.
+
+        Call update_lms_grades(self.context['request'], sequence=self.sequence, user_id=self.context['user_id'])
+        :return: nothing.
```
idegtiarov: L38 - 39 could be omitted they repeated method's body.

And instead of using code duplication I would rewrite send_grade in the base class as function with the parameter `with_request`
```
def _send_grade(self, with_request=False):
     is_send = with_request and self.context.get('request') or not self.context.get('request')
     if is_send:
         update_lms_grades(self.context.get('request'), sequence=self.sequence, user_id=self.context['user_id'])

@property
def send_grade(self):
     return _send_grade(self)
```

And use this property with the appropriate param in all policies.

alexbojko: I can't understand this line 
```
     is_send = with_request and self.context.get('request') or not self.context.get('request')
```

so I changed it to 
```
def _send_grade(self, with_request=False):
    if with_request:
        update_lms_grades(self.context.get('request'), sequence=self.sequence, user_id=self.context['user_id'])
    else:
        update_lms_grades(sequence=self.sequence, user_id=self.context['user_id'])
```

PR #105, opened 2018-02-20
Update VPAL engine recommend request with the optional LTI params.

idegtiarov:
Engine model is extended with the lti_parameters field
Sequence model is extended with the metadata JSON field add new
data which could be required by the engine driver.


bridge_adaptivity/module/models.py
```diff
@@ -232,6 +248,10 @@ def engine_driver(self):
             self.DRIVER = engine_driver
         return self.DRIVER
 
+    @property
```
alexbojko: as for me it looks like it should return dict, but it return tuple.

should it be better to rename this function to make it more clear?

alexbojko: for example 
```
def lti_params_list(...):
    ....
```
?

bridge_adaptivity/module/tests/test_vpal_engine.py
```diff
@@ -65,12 +65,16 @@ def test_fulfill_payload(self):
     @patch('requests.post', return_value=Mock(status_code=200))
     def test_select_activity(self, mock_post):
         expected_source_url = 'new_activity_source_url'
+        lti_param = "lis_person_sourcedid"
+        launch_params = {lti_param: "test_lis_person_sourcedid"}
+        self.sequence.fulfil_sequence_metadata(self.engine.lti_params, launch_params)
```
alexbojko: here's the same about renaming function... It's not clear why `lti_params` is list but `launch_params` is dict

bridge_adaptivity/module/models.py
idegtiarov: I don't want to have too long property name, btw property returns generator not list and it could be confusing expect list and get a generator. lti_params looks good to me.

bridge_adaptivity/module/tests/test_vpal_engine.py
idegtiarov: launch_params is the attribute of the ToolProvider and returns a dict, our logic require iterable object - generator in our case. All parameters look predictable to me.

PR #106, opened 2018-02-20
fix(AD-102): Don't delete collections when delete group, don't delete group when delete course.

alexbojko:
### [PROBLEM]
Is described in ticket AD-102

### [SOLUTION]

* set NULL in FK fields on delete 
* create migration


bridge_adaptivity/module/forms.py
```diff
@@ -27,7 +27,6 @@ class GroupForm(ModelForm):
         choices=GRADING_POLICY_CHOICES,
         required=True,
     )
-    course = forms.ModelChoiceField(queryset=Course.objects.all())
```
idegtiarov: I would leave this field in the form with the assumption it could left blank.

alexbojko: it's not needed, because `course` field in `CollectionGroup` model has `blank=True, null=True`

PR #109, opened 2018-02-21
fix(AD-109): When user update group - GetGradingPolicyForm view should pass policy to form.

alexbojko:
#### The problem is described in ticket AD-109.
The threshold is always zero after saving group setting

### [SOLUTION]
Add method `get_form_kwargs` to `GetGradingPolicyForm` view to pass instance policy to the form.


bridge_adaptivity/module/views.py
```diff
@@ -409,9 +420,7 @@ def callback_sequence_item_grade(request):
 
 
 def sync_collection(request, pk):
-    """
```
idegtiarov: why?

PR #110, opened 2018-02-21
Add filtering to the source activities on edit Collection view.

idegtiarov:


bridge_adaptivity/module/static/module/js/module.js
```diff
@@ -1,6 +1,22 @@
...
     $(function () {
         var defaultSourceItemTitle = "No Title";
+        var $filter = $("#filter"),
+            filterVal,
+            regexFilter;
+
+        $filter.change(function () {
+            filterVal = $("#filter").val();
+            if (filterVal) {
+                regexFilter = new RegExp(filterVal,"i");
```
alexbojko: add space after comma

bridge_adaptivity/module/static/module/js/module.js
```diff
@@ -128,6 +144,11 @@
                 } else {
                     sourceButton.text(item["display_name"]);
                 }
+                if (regexFilter) {
+                    if (!regexFilter.exec(sourceButton.text())) {
+                        listItem.detach()
```
alexbojko: if you want to use `.detach()` method you should store somewhere value, turned by this function, and then, when you will want to append it somewhere, you will use this value. (something like cache or memo var).

In this case, using `.detach()` is not needed, because once you call `detach()` and not assign returned value to some variable you loose this items, and browser will not know what to insert in future. So, here you can use just `.remove()`, because logic of this app is to build DOM every time. 

Here's example from api.jquery.com about `detach()` function:

```
var p;
$( "button" ).click(function() {
  if ( p ) {
    p.appendTo( "body" );
    p = null;
  } else {
    p = $( "p" ).detach();
  }
});
```

Aws you can see, they store all `p` tags in global variable when they do `detach()` and from other side `if(p)`  they append them to body.

So, no magic here.

bridge_adaptivity/module/static/module/js/module.js
idegtiarov: done

bridge_adaptivity/module/static/module/js/module.js
idegtiarov: updated with the `remove` method

PR #112, opened 2018-02-28
fix(AD-104): Added possibility to get courses from different content sources (not from only one).

alexbojko:
### [PROBLEM]
Is described in ticket [AD-104](https://vpalresearch.atlassian.net/browse/AD-104)

### [SOLUTION]
* remove `content_provider` field from `OAthClient` model
* add field `o_auth_client` into `LtiConsumer` model
* model `LtiConsumer` not inherit `ModelFieldIsDefaultMixin`, so it could be more than one active `content source`
* add migrations in `api` and `bridge_lti` with changes listed above
* make changes in `/api/sources` to handle `source_id` if there's more than one active `LtiConsumer` if only one - `source_id` parameter is not required
* changes in templates to pass `source_id` to JS handlers to pass it to `api`
* changes in `openedx.py` backend to be able to work with `multiple content sources`
* changes in JS to pass `content_source_id` to `api`


bridge_adaptivity/api/backends/openedx.py
```diff
@@ -117,71 +119,104 @@ def get_provider_courses(self, username=None, org=None, mobile=None):
...
-    except HttpNotFoundError:
-        raise HttpClientError(_("Requested course not found. Check `course_id` url encoding."))
-    except HttpClientError as exc:
-        raise HttpClientError(_("Not valid query: {}").format(exc.message))
+    for content_source in content_sources:
+        # Get API client instance:
+        api = OpenEdxApiClient(content_source=content_source)
+        try:
+            # filter function will be applied to api response
+            filter_func = partial(
```
idegtiarov: I would simplify the code with the explicit usage of the `apply_data_filter` function on the result of the  `api.get_course_blocks` function.

bridge_adaptivity/api/backends/openedx.py
```diff
@@ -117,71 +119,104 @@ def get_provider_courses(self, username=None, org=None, mobile=None):
...
+    :param source_id: LtiConsumer object id
+    :return: queryset of content_sources
+    :raise HttpClientError: if provided parameters are not valid
+    """
+    content_sources = get_content_providers(source_id=source_id)
+    if not content_sources:
+        # if no active content sources
+        raise HttpClientError(_("No active Content Provider"))
+
+    if not_allow_empty_source_id:
```
idegtiarov: I would move check for mandatory param to the top of the function to avoid odd database requests if the condition is True

bridge_adaptivity/api/backends/openedx.py
```diff
@@ -117,71 +119,104 @@ def get_provider_courses(self, username=None, org=None, mobile=None):
...
+    :param source_id: LtiConsumer id
     :return: (list) blocks data
     """
-    content_source = get_content_provider()
-    if not content_source:
-        raise HttpClientError(_("No active Content Provider"))
-
-    # Get API client instance:
-    api = OpenEdxApiClient(content_source=content_source)
+    content_sources = check_source_id(source_id)
```
idegtiarov: We did API call for available blocks every time we change the course, why do we need to get through all content sources if we are interested in the certain source and course?

bridge_adaptivity/api/backends/openedx.py
```diff
@@ -117,71 +119,104 @@ def get_provider_courses(self, username=None, org=None, mobile=None):
...
-    try:
-        courses_list = api.get_provider_courses()
-        filtered_courses = apply_data_filter(courses_list, filters=['id', 'course_id', 'name', 'org'])
-    except HttpClientError as exc:
-        raise HttpClientError(_("Not valid query: {}").format(exc.message))
+    for content_source in content_sources:
+        # Get API client instance:
+        api = OpenEdxApiClient(content_source=content_source)
+        try:
+            filter_func = partial(
```
idegtiarov: ditto as in L166

bridge_adaptivity/api/backends/openedx.py
```diff
@@ -190,43 +225,43 @@ def apply_data_filter(data, filters=None, context_id=None):
...
 
 
-def get_content_provider():
+def get_content_providers(source_id=None):
     """
     Pick active (enabled) content Sources (aka Providers).
 
     LTI Providers which expose courses content blocks to use in adaptive collections.
-    :return: content_source
+    :param source_id: LtiConsumer ID or None
```
idegtiarov: `or None` is useless here

bridge_adaptivity/bridge_lti/consumer.py
```diff
@@ -36,9 +36,15 @@ def source_preview(request):
...
-    content_provider = get_content_provider()
-    if not content_provider:
+    content_source_id = request.GET.get('content_source_id')
+    content_providers = get_content_providers(content_source_id)
+    require_content_source_id = content_providers.count() > 1
+
+    if require_content_source_id and not content_source_id:
         return render(request, 'bridge_lti/stub.html')
+
+    content_provider = content_providers.first()
```
idegtiarov: It seems this part is not finished yet. Why we take the first of content_providers what about others?

bridge_adaptivity/bridge_lti/models.py
```diff
@@ -32,19 +31,19 @@ def __str__(self):
...
     Content source connections.
     """
 
-    IS_DEFAULT_FIELD = 'is_active'
     name = fields.CharField(max_length=255, blank=True, null=True, unique=True)
     provider_key = models.CharField(max_length=255)
     provider_secret = models.CharField(max_length=255)
     lti_metadata = fields.CharField(max_length=255, null=True, blank=True)
     host_url = models.URLField(max_length=255, null=True)
+    o_auth_client = models.ForeignKey('api.OAuthClient', null=True)
```
idegtiarov: I would set this field to be mandatory and not empty to avoid creating Consumer without oauth_client.

bridge_adaptivity/bridge_lti/models.py
```diff
@@ -32,19 +31,19 @@ def __str__(self):
 
 
 @python_2_unicode_compatible
-class LtiConsumer(ModelFieldIsDefaultMixin, models.Model):
+class LtiConsumer(models.Model):
```
idegtiarov: We remove IsDefault mixin, what about the case with no active source, I would add the new mixing with the check there is at least one active source and point admin if he trying to switch off all sources.

bridge_adaptivity/module/static/module/js/module.js
```diff
@@ -180,15 +190,18 @@
                 e.stopImmediatePropagation();
                 $("#sourceModal").modal("show");
                 $("#sourceModalLabel").text(title);
-                configurePreview(title, ltiUrl, sourceId, modalFrame);
+                configurePreview(this, title, ltiUrl, sourceId, contentSourceId, modalFrame);
             })
         }
 
-        function configurePreview(title, ltiUrl, sourceId, modalFrame) {
+        function configurePreview(me, title, ltiUrl, sourceId, contentSourceId, modalFrame) {
```
idegtiarov: `me` is not used and could be omitted.

bridge_adaptivity/module/static/module/js/module.js
```diff
@@ -180,15 +190,18 @@
...
             var idParam = "source_id=" + sourceId + "&";
             var displayNameParam = "source_name=" + title + "&";
             var ltiUrlParam = "source_lti_url=" + ltiUrl + "&";
-            var previewUrl = internalUrls.ltiSourcePreview + "?" + idParam + displayNameParam + ltiUrlParam;
+            var contentSourceIdParam = "content_source_id=" + contentSourceId + "&"
```
idegtiarov: missed trailing semicolumn.

bridge_adaptivity/module/views.py
```diff
@@ -225,8 +218,8 @@ def get_success_url(self):
 @method_decorator(login_required, name='dispatch')
 class ActivityUpdate(CollectionIdToContextMixin, UpdateView):
     model = Activity
+    form_class = ActivityCreate.form_class
```
idegtiarov: Please use explicit `ActivityForm` instead inheriting ActivityCreate class.

bridge_adaptivity/bridge_lti/consumer.py
alexbojko: I already explained it to you. 
because here we will have only one in queryset.  

Probably I should better to add comment here.

bridge_adaptivity/bridge_lti/models.py
alexbojko: it was not possible, because migration will not work. `null=True` is ok, because form will require this field because it has `blank=False`.

bridge_adaptivity/bridge_lti/models.py
alexbojko: make sense

bridge_adaptivity/bridge_lti/models.py
idegtiarov: Make sense, thanks for the explanation.

PR #118, opened 2018-03-12
fix(AD-117): Add group in course functionality inplementation.

alexbojko:
### [PROBLEM]
is described in ticket AD-117

### [SOLUTION]
* Add modal with form on course detail page to add groupS into this course


bridge_adaptivity/module/forms.py
```diff
@@ -75,3 +75,21 @@ class Meta:
         widgets = {
             'name': forms.HiddenInput(),
         }
+
+
+class AddCourseGroupForm(forms.Form):
+    def __init__(self, *args, **kwargs):
+        user = kwargs.pop('user')
+        super(AddCourseGroupForm, self).__init__(*args, **kwargs)
+        if user:
```
idegtiarov: L84 looks redundant in case L82 doesn't have a default value.

PR #133, opened 2018-05-25
Move to django2

flying-pi:


bridge_adaptivity/config/urls.py
```diff
@@ -1,26 +1,29 @@
...
+from module.urls import urlpatterns as module
 from . import views
 
 
 urlpatterns = [
-    url(r'^login/$', auth_views.login, name='login'),
-    url(r'^logout/$', auth_views.logout, name='logout'),
-    url(r'^admin/', include(admin.site.urls)),
-    url(r'^health/$', views.health),
+    path(r'login/', auth_views.login, name='login'),
```
idegtiarov: regular expressions are not used with the `path` method

PR #141, opened 2018-06-06
Add LTI component as the Activity to the Bridge's Collection

idegtiarov:
Remove filters from the types of received blocks.
Small style fixes in the module.js file.


bridge_adaptivity/module/static/module/js/module.js
```diff
@@ -28,12 +28,18 @@
...
             }
         });
-        var typeToIcon = {
-            html: "glyphicon-list-alt",
-            problem: "glyphicon-question-sign",
-            video: "glyphicon-facetime-video",
-            other: "glyphicon-option-horizontal"
-        };
+
+        function typeToIcon(itemType) {
```
flying-pi: please use switch construction

bridge_adaptivity/module/static/module/js/module.js
```diff
@@ -131,7 +137,7 @@
                 $("<span/>")
                     .data("toggle", "tooltip")
                     .attr("title", item.type)
-                    .addClass("badge pull-left glyphicon " + typeToIcon[item.type])
+                    .addClass("badge pull-left glyphicon " + typeToIcon(item.type))
```
flying-pi: Maybe template strings more suitable?

bridge_adaptivity/module/static/module/js/module.js
idegtiarov: Nice catch, done.

bridge_adaptivity/module/static/module/js/module.js
idegtiarov: Yep, template string would be better to use.

PR #145, opened 2018-06-18
Added tests for adding and removing course for the group.

flying-pi:
Added tests for adding and removing course for the group.
Added optional for generation coverage report


bridge_adaptivity/module/tests/test_views.py
```diff
@@ -72,10 +66,36 @@ def setUp(self, mock_apply_async):
             consumer_name='consumer_name',
             consumer_key='consumer_key',
             consumer_secret='consumer_secret',
-            expiration_date=datetime.datetime.today() + timedelta(days=1),
+            expiration_date=datetime.datetime.today() + datetime.timedelta(days=1),
             lms_metadata='lms_metadata'
         )
 
+    def create_grading_policy(self, name=None, public_name=None, threshold=None, engine=None, is_default=True):
+        name = name or 'trials_count'
```
idegtiarov: why do you choose trails_count as default grading policy?

bridge_adaptivity/module/tests/test_views.py
```diff
@@ -184,6 +204,32 @@ def test_update_cg(self):
         self.assertNotEqual(test_g.description, self.test_cg.description)
         self.assertNotEqual(test_g.collections.all(), self.test_cg.collections.all())
 
+    def test_add_group_to_course(self):
+        group = self.create_group()
```
idegtiarov: I would move this functions create_group and create_course to the setup class instead of using as separate functions.

PR #150, opened 2018-06-25
Add sequence preview 

flying-pi:


bridge_adaptivity/module/views.py
```diff
@@ -525,3 +525,22 @@ def sync_collection(request, slug):
...
     )
     return redirect(reverse('module:collection-detail', kwargs={'slug': slug}) + '?back_url={}'.format(back_url))
+
+
+def preview_collection(request, slug):
+    acitvities = [
+        {
+            'url': f'{reverse("lti:source-preview")}?source_id={a.id}&source_name={a.name}&source_lti_url='
+                   f'{a.source_launch_url}&content_source_id={a.lti_consumer_id}',
+            'pos': pos + 1,
```
idegtiarov: I would use `start` param in the enumerate function instead of adding 1.

PR #154, opened 2018-07-05
Feature manually update students grades

idegtiarov:
Currently, if the grading policy is updated in the collection group it will affect only students who are still working with the sequences related to the group.

The new feature makes it possible for the instructor to update all students grades based on the sequences related to the group.


bridge_adaptivity/module/views.py
```diff
@@ -185,9 +185,8 @@ def get_has_available_objects(self, form):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        context.update({
-            'bridge_host': settings.BRIDGE_HOST,
-        })
+        context['bridge_host'] = settings.BRIDGE_HOST
```
flying-pi: Why you avoid `update` function?

bridge_adaptivity/module/views.py
```diff
@@ -527,6 +524,17 @@ def sync_collection(request, slug):
     return redirect(reverse('module:collection-detail', kwargs={'slug': slug}) + '?back_url={}'.format(back_url))
 
 
+def update_students_grades(request, group_slug):
```
flying-pi: For me more suitable use jquery for this operation

idegtiarov: Thanks, nice catch I agree.

idegtiarov: But you know it would require additional conditions or code duplication because we use the same modal template for delete, sync collections, and students grade update operation and I suppose this change is not required now.

PR #159, opened 2018-07-06
Add collection creation modal in the group

flying-pi:
Add modals for the creation or updating several items


bridge_adaptivity/module/mixins/views.py
```diff
@@ -138,7 +138,8 @@ class ModalFormMixin(object):
 
     def form_valid(self, form):
         super().form_valid(form)
-        return JsonResponse(status=200, data={'status': 'ok'})
+        # Return status code as Accepted and json {'status': 'ok}
```
idegtiarov: I would move and extend comment as method description.

bridge_adaptivity/module/static/module/js/module.js
```diff
@@ -363,7 +355,7 @@
                     success: modalSuccessListener(url)
                 }
             );
-            return false;
+            // return false;
```
idegtiarov: why is this line commented?

bridge_adaptivity/module/urls.py
```diff
@@ -20,7 +20,7 @@
         name='rm-group-from-course'),
 
     url(r'^group/$', GroupList.as_view(), name='group-list'),
-    url(r'^group/add/?$', GroupCreate.as_view(), name='group-add'),
+    url(r'^(?:course/(?P<course_slug>[\w-]+)/)?group/add/?$', GroupCreate.as_view(), name='group-add'),
```
idegtiarov: I would use `path` instead of `url` in all new changes to the urls.py file.

bridge_adaptivity/module/views.py
```diff
@@ -228,7 +233,12 @@ class CollectionList(BaseCollectionView, ListView):
 
 @method_decorator(login_required, name='dispatch')
 class CollectionCreate(BaseCollectionView, SetUserInFormMixin, ModalFormMixin, CreateView):
-    pass
+
+    def form_valid(self, form):
+        result = super().form_valid(form)
+        if 'group_slug' in self.kwargs and self.kwargs['group_slug']:
```
idegtiarov: I would use the `get` method instead of having two conditional statements.

bridge_adaptivity/module/views.py
```diff
@@ -305,26 +315,38 @@ def get_object(self, queryset=None):
...
     model = Activity
     form_class = ActivityForm
 
+    def get_initial(self):
+        result = super().get_initial()
+        if self.request.method == 'GET':
+            result.update({
+                'name': self.request.GET.get('name'),
+                'source_name': self.request.GET.get('source_name'),
+                'source_launch_url': self.request.GET.get('source_launch_url').replace(' ', '+'),
```
idegtiarov: here and lower what about the case when `get` returns None?

bridge_adaptivity/module/views.py
```diff
@@ -305,26 +315,38 @@ def get_object(self, queryset=None):
...
-class ActivityUpdate(CollectionSlugToContextMixin, UpdateView):
+class ActivityUpdate(CollectionSlugToContextMixin, ModalFormMixin, UpdateView):
     model = Activity
     form_class = ActivityForm
     context_object_name = 'activity'
 
     def get(self, request, *args, **kwargs):
         activity = self.get_object()
-        if 'direction' in kwargs:
+        if 'direction' in kwargs and kwargs['direction']:
```
idegtiarov: ditto, `get` method will suit better for this statement instead of two checks.

PR #162, opened 2018-07-09
Add message for prevent preview for the empty group

flying-pi:


bridge_adaptivity/module/static/module/js/module.js
```diff
@@ -365,5 +365,13 @@
             );
             return false;
         });
+
+        $('#collection-preview-button').click(e => {
+            if($('.activity').length === 0) {
+                $('#no-preview-warning').modal('show');
```
idegtiarov: Please rename `id` to some implicit value

PR #165, opened 2018-07-11
Fix creating activity with empty stype.

idegtiarov:


idegtiarov:
Better for what purpose?

PR #173, opened 2018-08-20
Mark as problem only for problem

flying-pi:
Resolve https://vpalresearch.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=AD&modal=detail&selectedIssue=AD-215


idegtiarov:
It has become empty.

PR #177, opened 2018-08-21
Fix sequence is completed marker

idegtiarov:
Recently, if the first choosing activity for the sequence fails sequence
is marked as completed. Now sequence marked as completed only on the
certain response from the engine.

### DIFF
--- a/bridge_adaptivity/module/utils.py
+++ b/bridge_adaptivity/module/utils.py
@@ -1,16 +1,28 @@
-from django.http import Http404
-from django.shortcuts import get_object_or_404
+from logging import getLogger
 
 from module.models import Activity
+
+log = getLogger(__name__)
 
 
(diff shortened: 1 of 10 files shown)


bridge_adaptivity/module/utils.py
```diff
@@ -1,16 +1,28 @@
...
+        activity_source_launch_url = engine_choose.get('source_launch_url')
+        if engine_choose.get('complete'):
+            sequence.completed = True
+            sequence.save()
+            return
+        elif activity_source_launch_url:
+            return Activity.objects.filter(
+                collection=sequence.collection, source_launch_url=activity_source_launch_url
+            ).first()
+    except AttributeError:
```
flying-pi: In which place this exception can happen?

idegtiarov: It may happen if engine recommendation returns None object.

PR #180, opened 2018-08-23
Fix sequence completed strict forward

idegtiarov:
The fix is required if non_problem activity appears first in the sequence and page is refreshed.


bridge_adaptivity/bridge_lti/provider.py
```diff
@@ -144,17 +144,17 @@ def create_sequence_item(request, sequence, start_activity, tool_provider, lti_c
     return sequence_item
 
 
-def announcement_page(request):
+def stub_page(request, title=None, message=None, tip=None):
```
flying-pi: Please seat meaningful value for default arguments instead None

idegtiarov: Nice catch, thanks!

PR #181, opened 2018-08-30
Sequence item visibility (fix for AD-222 and AD-225)

idegtiarov:
Recently, in a strict forward case for collection with non-problem activity, sequence items were shown in an incorrect way after the page was refreshed.

### DIFF
--- a/bridge_adaptivity/module/views.py
+++ b/bridge_adaptivity/module/views.py
@@ -18,6 +18,7 @@ from lti.outcome_response import CODE_MAJOR_CODES, SEVERITY_CODES
 from slumber.exceptions import HttpClientError
 
 from api.backends.api_client import get_active_content_sources, get_available_courses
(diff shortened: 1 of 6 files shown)


bridge_adaptivity/module/views.py
```diff
@@ -392,13 +393,19 @@ class SequenceItemDetail(LtiSessionMixin, DetailView):
...
-            item_filter.update({'score__isnull': False})
+        sequence_items = SequenceItem.objects.filter(sequence=self.object.sequence)
+        last_item = sequence_items.last()
+        if (
+            self.request.session.get('Lti_update_activity') and
+            len(sequence_items) > 1 and
+            last_item.is_problem and
+            not last_item.score
+        ):
+            sequence_items = sequence_items[:len(sequence_items) - 1]
```
flying-pi: You can use negative index here

PR #202, opened 2018-12-13
Config update

kunanit:
Some small configuration updates:
* Depreciate entrypoint script and simply dockerfile
* gitignore some more file types
* add dockerignore file to optimize docker image builds


kunanit:
@idegtiarov Doesn't look like `entrypoint.sh` or `Dockerfile` was being used by the RG stage deploy script or the local docker-compose config, but let me know if any of these changes might impact you

idegtiarov:
@kunanit You are right we do not use these files on the RG side.

PR #212, opened 2019-01-29
Reduce db queries in Log admin views

kunanit:
Optimize the Log admin views by reducing the number of database queries made. The sequence_item field in the detail view is made read-only, to prevent a potentially large dropdown menu from being populated (can cause problems loading if there are many sequence items).


idegtiarov:
Nice catch!

PR #224, opened 2019-02-27
Optimization alert related to show errors

AndreyLykhoman:
*Description*
Optimization alert related to show errors with content courses.
*Tasks*
https://vpalresearch.atlassian.net/browse/AD-270
https://youtrack.raccoongang.com/issue/ALOSI-42

### DIFF
--- a/bridge_adaptivity/api/backends/api_client.py
+++ b/bridge_adaptivity/api/backends/api_client.py
@@ -96,7 +96,7 @@ def get_available_courses(request, source_ids=None):
     content_sources = get_active_content_sources(request, source_ids, not_allow_empty_source_id=False)
 
     all_courses = []
-    errors = []
+    errors = {}
     for content_source in content_sources:
         # Get API client instance:
         try:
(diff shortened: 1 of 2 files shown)


bridge_adaptivity/api/backends/api_client.py
```diff
@@ -109,12 +109,12 @@ def get_available_courses(request, source_ids=None):
                 )
             )
         except HttpClientError as exc:
-            errors.append(
-                _(
-                    "Content source '{}' is unavailable. Reason: {} Please contact Bridge administrator to solve the"
-                    " problem."
-                ).format(content_source.name, str(exc))
-            )
+            error_list = list()
```
idegtiarov: I would use literal syntax for creating new objects.

bridge_adaptivity/api/backends/api_client.py
```diff
@@ -109,12 +109,12 @@ def get_available_courses(request, source_ids=None):
...
+            if str(exc) in errors:
```
idegtiarov: Creating a new variable with value `str(exc)` suits better for cases when we use it more than one time.

bridge_adaptivity/api/backends/api_client.py
```diff
@@ -109,12 +109,12 @@ def get_available_courses(request, source_ids=None):
...
+                error_list = errors[str(exc)]
+            error_list.append(content_source.name)
```
idegtiarov: I would use `defaultdict` instead of writing so long construction

PR #231, opened 2019-04-09
Fix step, max and min limit parameters for difficulty field

AndreyLykhoman:
Fix step, max, and min limit parameters for difficulty field

https://youtrack.raccoongang.com/issue/ALOSI-37
https://vpalresearch.atlassian.net/browse/AD-271

### DIFF
--- a/bridge_adaptivity/module/forms.py
+++ b/bridge_adaptivity/module/forms.py
@@ -30,7 +30,10 @@ class ActivityForm(ModelForm):
         model = Activity
         exclude = ['collection', 'points']
         widgets = {
-            'stype': forms.HiddenInput(), 'points': forms.HiddenInput(), 'lti_lms_platform': forms.HiddenInput()
+            'stype': forms.HiddenInput(),
+            'points': forms.HiddenInput(),
+            'lti_lms_platform': forms.HiddenInput(),
+            'difficulty': forms.NumberInput(attrs={'step': "0.1", "min": 0.0, "max": 1.0})
         }
 
 
--- a/bridge_adaptivity/module/forms.py
+++ b/bridge_adaptivity/module/forms.py
@@ -32,7 +32,7 @@ class ActivityForm(ModelForm):
         widgets = {
             'stype': forms.HiddenInput(),
             'lti_lms_platform': forms.HiddenInput(),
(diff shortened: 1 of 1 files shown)


bridge_adaptivity/module/forms.py
```diff
@@ -30,7 +30,10 @@ class Meta:
         model = Activity
         exclude = ['collection', 'points']
         widgets = {
-            'stype': forms.HiddenInput(), 'points': forms.HiddenInput(), 'lti_lms_platform': forms.HiddenInput()
+            'stype': forms.HiddenInput(),
+            'points': forms.HiddenInput(),
+            'lti_lms_platform': forms.HiddenInput(),
+            'difficulty': forms.NumberInput(attrs={'step': "0.1", "min": 0.0, "max": 1.0})
```
arsentur: We can add a trailing comma. This will make a smaller diff for the following changes. But it's up to you.

bridge_adaptivity/module/forms.py
```diff
@@ -30,7 +30,10 @@ class Meta:
         model = Activity
         exclude = ['collection', 'points']
         widgets = {
-            'stype': forms.HiddenInput(), 'points': forms.HiddenInput(), 'lti_lms_platform': forms.HiddenInput()
+            'stype': forms.HiddenInput(),
```
arsentur: Why do we need hidden inputs in the form?

idegtiarov: Why does lti_lms_platform is removed from hidden input?

bridge_adaptivity/module/forms.py
```diff
@@ -30,7 +30,10 @@ class Meta:
...
+            'points': forms.HiddenInput(),
```
idegtiarov: `points` are excluded from the form do we need to add hiddenInput for this field?

bridge_adaptivity/module/forms.py
AndreyLykhoman: You are right. I fixed this.

bridge_adaptivity/module/forms.py
AndreyLykhoman: You are right. I deleted this field.

PR #232, opened 2019-04-22
Using the Django channels for the button "Next" in the demo and the learner flow.

AndreyLykhoman:
Description:  Using Django channels for the button "Next" in the demo and the learner flow.
Using Redis like CHANNEL_LAYERS.

Tasks: 
https://vpalresearch.atlassian.net/browse/AD-267
https://youtrack.raccoongang.com/issue/ALOSI-76


bridge_adaptivity/Dockerfile_local
```diff
@@ -7,3 +7,27 @@ RUN mkdir /bridge_adaptivity
...
+ENV CHROMEDRIVER_DIR /chromedriver
+RUN mkdir $CHROMEDRIVER_DIR
+
+# Download and install Chromedriver
+RUN wget -q --continue -P $CHROMEDRIVER_DIR "http://chromedriver.storage.googleapis.com/$CHROMEDRIVER_VERSION/chromedriver_linux64.zip"
+RUN unzip $CHROMEDRIVER_DIR/chromedriver* -d $CHROMEDRIVER_DIR
+
+# Put Chromedriver into the PATH
+ENV PATH $CHROMEDRIVER_DIR:$PATH
+RUN chmod +x /chromedriver
```
idegtiarov: Andrey, please be attentive! ^^

bridge_adaptivity/config/routing.py
```diff
@@ -0,0 +1,11 @@
+from channels.auth import AuthMiddlewareStack
+from channels.routing import ProtocolTypeRouter, URLRouter
+import module.routing
```
idegtiarov: local packages are the third group of imports, please separate it from the second one. 

BTW I would import `websocket_urlpatterns` from the `module.routing`

bridge_adaptivity/config/settings/base.py
```diff
@@ -192,3 +193,14 @@
 
 # Demo functionality
 TEST_SEQUENCE_SUFFIX = 'test_sequence_suffix'
+
```
idegtiarov: I would add some comments about the following configurations.

bridge_adaptivity/module/consumers.py
```diff
@@ -0,0 +1,44 @@
+from asgiref.sync import async_to_sync
+import channels.layers
+from channels.generic.websocket import AsyncWebsocketConsumer
+import json
```
idegtiarov: Import groups are improperly configured, please improve.

bridge_adaptivity/module/consumers.py
```diff
@@ -0,0 +1,44 @@
...
+
+class NextButtonConsumer(AsyncWebsocketConsumer):
+
+    prefix = 'activity'
+
+    async def connect(self):
+        self.room_name = f"{self.scope['url_route']['kwargs']['room_name']}"
```
idegtiarov: Room? Why do we need such odd naming?

bridge_adaptivity/module/consumers.py
```diff
@@ -0,0 +1,44 @@
...
+        self.room_group_name = f'{self.prefix}_{self.room_name}'
```
idegtiarov: ditto
BTW I do not think we need to set the `prefix` as a class attribute

bridge_adaptivity/module/consumers.py
```diff
@@ -0,0 +1,44 @@
...
+
+        # Join room group
```
idegtiarov: ditto

bridge_adaptivity/module/consumers.py
```diff
@@ -0,0 +1,44 @@
...
+        await self.accept()
+
+    async def disconnect(self, close_code):
+        # Leave room group
+        await self.channel_layer.group_discard(
+            self.room_group_name,
+            self.channel_name
+        )
+
+    async def receive(self, text_data):
```
idegtiarov: why do we need an empty method?

bridge_adaptivity/module/consumers.py
```diff
@@ -0,0 +1,44 @@
...
+
+        # Join room group
+        await self.channel_layer.group_add(
+            self.room_group_name,
+            self.channel_name
+        )
+        await self.accept()
+
+    async def disconnect(self, close_code):
+        # Leave room group
```
idegtiarov: ditto

bridge_adaptivity/module/consumers.py
```diff
@@ -0,0 +1,44 @@
...
+        await self.channel_layer.group_discard(
+            self.room_group_name,
+            self.channel_name
+        )
+
+    async def receive(self, text_data):
+        pass
+
+    async def chat_message(self, event):
```
idegtiarov: `chat`?  what chat?

bridge_adaptivity/module/consumers.py
```diff
@@ -0,0 +1,44 @@
...
+
+        # Send message to WebSocket
+        await self.send(text_data=json.dumps({
+            'result': message
+        }))
+
+    @staticmethod
+    def send_message_to_channel(room_name, message):
+        channel_layer = channels.layers.get_channel_layer()
+        async_to_sync(channel_layer.group_send)(
```
idegtiarov: `async_to_sync` could you please clarify why do we need this step back to the sync paradigma?

bridge_adaptivity/module/routing.py
```diff
@@ -0,0 +1,8 @@
+# chat/routing.py
```
idegtiarov: `chat` again, please clear the code from the odd artifacts, so we can continue reviewing process.

.travis.yml
```diff
@@ -1,23 +1,36 @@
...
-- linux
-sudo: false
-language: python
-python:
-- "3.6"
-install: pip install tox-travis
-script: tox
+jobs:
+  include:
+    - stage: flake8
```
idegtiarov: Cannot find selenium stage, did we add it into the CI process?

update_stage.sh
```diff
@@ -7,4 +7,5 @@ git pull
 
 cd ./bridge_adaptivity
 
-docker-compose -f ./docker-compose-stage.yml restart
+make migrate
```
idegtiarov: It looks that it should be `migrate-stage`.

update_stage.sh
```diff
@@ -7,4 +7,5 @@ git pull
...
+make run-stage
```
idegtiarov: run-stage wrapping `up` command which is not the same as `restart`, one I missed something.

.travis.yml
AndreyLykhoman: We will add this when selenium tests are written.

update_stage.sh
AndreyLykhoman: You are right.

update_stage.sh
AndreyLykhoman: This command can recreate containers for a service if it finds changes in images or configurations (you can read more here https://docs.docker.com/compose/reference/up/). 
And this we can use as restart containers that have changed (instead of using restarting command that stop and recreate all containers).

idegtiarov:
@AndreyLykhoman please resolve conflicts.

update_stage.sh
idegtiarov: Thanks for the clarification

PR #243, opened 2019-05-29
Reorganization activity form

AndreyLykhoman:
Description:  Changes:
 - Add lti_content_source to advanced fields.
 - Add  js script for changing Show/Hide advanced options button

Tasks: 
https://youtrack.raccoongang.com/issue/ALOSI-94
https://youtrack.raccoongang.com/issue/ALOSI-95


bridge_adaptivity/module/templates/module/activity_modal_form.html
```diff
@@ -27,15 +27,15 @@
         <button
                 class="btn btn-default activity-show-advanced-options"
                 data-toggle=".advanced_fields"
-                onclick="$($(this).data('toggle')).toggle('slow'); event.preventDefault();">
```
idegtiarov: I would update `onclick` button's parameter instead. Why it won't work, what do you think?

AndreyLykhoman: I  rebased js code into `onclick` parameter.

idegtiarov: Cool, thanks

PR #252, opened 2019-09-18
Fix js mistakes and change postgres version and  psycopg2 version

AndreyLykhoman:
Description:  This commit has:
 - Change version of psycopg2, and write the version of Postgres in the docker-compose file.
 - Fix grading_policy_form URL. 
 - Fix js script for reloading a page when a form already sent.

Main Task:   https://youtrack.raccoongang.com/issue/ALOSI-109
Task for review: https://youtrack.raccoongang.com/issue/ALOSI-112

### DIFF
--- a/bridge_adaptivity/module/static/module/js/module.js
+++ b/bridge_adaptivity/module/static/module/js/module.js
@@ -360,9 +360,14 @@
         const modalSuccessListener = (url) => {
             return result => {
                 let modal = $('#modal-wrapper');
-                if (result.status !== undefined && result.status === 'ok' && result.collection_slug !== undefined) {
-                    var new_url = new URL(window.location.href);
+                let new_url = new URL(window.location.href);
+
+                // FixMe: Following statement checks 'collection_slug' in the result and adds get params to the URL.
+                //        This is required to have correct redirection in case collection slug is changed.
+                if (result.collection_slug !== undefined) {
                     new_url.searchParams.append('collection_slug', result.collection_slug);
+                }
+                if (result.status === 'ok') {
                     window.location.href = new_url.href;
                     return;
                 }
--- a/bridge_adaptivity/docker-compose.yml
+++ b/bridge_adaptivity/docker-compose.yml
@@ -4,7 +4,7 @@ version: '3.4'
 services:
   postgres:
     container_name: postgresql_BFA
-    image: postgres
+    image: postgres:11.5
     env_file: ./envs/pg.env
     volumes:
       - pgs:/var/lib/postgresql/data/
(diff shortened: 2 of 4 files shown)


bridge_adaptivity/module/static/module/js/module.js
```diff
@@ -360,9 +360,12 @@
         const modalSuccessListener = (url) => {
             return result => {
                 let modal = $('#modal-wrapper');
-                if (result.status !== undefined && result.status === 'ok' && result.collection_slug !== undefined) {
-                    var new_url = new URL(window.location.href);
+                var new_url = new URL(window.location.href);
+                 // FixMe: This check  is needed to remove in another place
```
idegtiarov: Please rewrite the comment id doesn't clear enough.

bridge_adaptivity/module/static/module/js/module.js
```diff
@@ -360,9 +360,12 @@
         const modalSuccessListener = (url) => {
             return result => {
                 let modal = $('#modal-wrapper');
-                if (result.status !== undefined && result.status === 'ok' && result.collection_slug !== undefined) {
-                    var new_url = new URL(window.location.href);
+                var new_url = new URL(window.location.href);
```
idegtiarov: `let` definition for the `new_url` variable will be faire enough.

bridge_adaptivity/module/static/module/js/module.js
```diff
@@ -360,9 +360,12 @@
...
+                 // FixMe: This check  is needed to remove in another place
+                if (result.collection_slug !== undefined) {
                     new_url.searchParams.append('collection_slug', result.collection_slug);
+                }
+                if (result.status !== undefined && result.status === 'ok') {
```
idegtiarov: Why do we need two statements it seems that second covers both cases.

bridge_adaptivity/module/static/module/js/module.js
AndreyLykhoman: Done

bridge_adaptivity/module/static/module/js/module.js
AndreyLykhoman: Done

bridge_adaptivity/module/static/module/js/module.js
AndreyLykhoman: The first statement is used when the collection id changed. And it's used by only one flow: When we use CollectionUpdate view
The second state is used by all flows to create or update something. When a form saved correctly, we could reload the page.

idegtiarov: I meant `result.status !== undefine` is totally the same as we check `result.status === 'ok'`. So why do we need the first check?

AndreyLykhoman: You are right! We don't need the first check. I fixed this.

PR #253, opened 2019-09-20
[ALOSI] Round score to 3 places

AndreyLykhoman:
**Description**:  Rounds the value down to 3 decimal places. Displays the student the result in the form of a percent.
**Main Task**: https://youtrack.raccoongang.com/issue/ALOSI-116
**Task for review**: https://youtrack.raccoongang.com/issue/ALOSI-127

### DIFF
--- a/bridge_adaptivity/module/models.py
+++ b/bridge_adaptivity/module/models.py
@@ -2,6 +2,7 @@ import hashlib
 import importlib
 import inspect
 import logging
+import math
 import os
 import uuid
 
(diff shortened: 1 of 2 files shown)


bridge_adaptivity/module/models.py
```diff
@@ -126,7 +127,9 @@ def sequence_ui_details(self):
                 )
             elif ui_option == CollectionOrder.OPTIONS[1][0]:
                 grade = self.collection_order.grading_policy.calculate_grade(self)
-                details = f"{CollectionOrder.OPTIONS[1][1]}: {grade}"
+                if isinstance(grade, float):
```
idegtiarov: Please clarify why do we check float type here?

bridge_adaptivity/module/models.py
```diff
@@ -126,7 +127,9 @@ def sequence_ui_details(self):
...
+                    grade = round(grade * 100, 1)
```
idegtiarov: It seems we've already cut useless part of the number, why do we need round here?

bridge_adaptivity/module/models.py
```diff
@@ -2,6 +2,7 @@
 import importlib
 import inspect
 import logging
+from math import floor
```
idegtiarov: I would prefer to import library not function to ensure we do not override something important.

PR #254, opened 2019-10-02
[ALOSI] Add showing pop-up with congratulation messages if score 95% or above

AndreyLykhoman:
**Description:** Show pop-up with congratulation messages if users score 95% or above. I used a Web Socket.  And  I used Cookies to show pop-up message once if users score equal/above 95%.  A user can show this pop-up another one time If users score decrease(a cookie is deleted) and the next step increase (equal/above 95%). 
**Task:** https://youtrack.raccoongang.com/issue/ALOSI-119


bridge_adaptivity/module/views.py
```diff
@@ -745,6 +743,23 @@ def callback_sequence_item_grade(request):
     return HttpResponse(xml, content_type="application/xml")
 
 
+def _check_and_build_web_socket_message(sequence, score):
```
AndreyLykhoman: I made this function because, the Code Climate alarmed the error related to `Cyclomatic complexity is too high in function callback_sequence_item_grade. (11)`

bridge_adaptivity/module/static/module/js/forbid_next.js
```diff
@@ -1,54 +1,65 @@
...
+    uiDetailsArray.forEach(function(ui_option) {
+        let new_element = document.createElement(child_elements);
         new_element.className = child_classes;
         new_element.innerHTML = ui_option;
         ui_option_element.append(new_element);
     })
-};
+}
+
+function show_congratulation_pop_up(congratulation_cookie_name){
```
idegtiarov: Why did you decide to use cookies to show the message

bridge_adaptivity/module/static/module/js/work_with_cookie.js
```diff
@@ -0,0 +1,12 @@
+function getCookie(name) {
```
idegtiarov: Please add more comments. For newly created functions here and elsewhere.

bridge_adaptivity/module/static/module/js/work_with_cookie.js
```diff
@@ -0,0 +1,12 @@
...
+    let v = document.cookie.match('(^|;) ?' + name + '=([^;]*)(;|$)');
+    return v ? v[2] : null;
```
idegtiarov: Clarify what is v[2] param related to?

bridge_adaptivity/module/static/module/js/work_with_cookie.js
```diff
@@ -0,0 +1,12 @@
...
+}
+
+function setCookie(name, value, minutes) {
+    let d = new Date;
+    d.setTime(d.getTime() + 60*1000*minutes);
```
idegtiarov: Please add more spaces.

bridge_adaptivity/module/templates/module/modals/congratulation_message_modal.html
```diff
@@ -0,0 +1,18 @@
...
+<div class="modal fade" id="congratulationMessage" role="dialog">
+  <div class="modal-dialog" role="dialog">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span
+          aria-hidden="true">&times;</span></button>
+          <h3 class="modal-title">Congratulations!</i></h3>
+      </div>
+      <div class="modal-body">
+          <h4>You have achieved an excellent score on this assignment. This is a good time to move on to the next part of the course.</h4>
```
idegtiarov: I would prefer we let the instructor update Congratulation text.

bridge_adaptivity/module/views.py
idegtiarov: Usually, when the new function is created it is placed upper the place where it is used.

bridge_adaptivity/module/views.py
```diff
@@ -745,6 +743,23 @@ def callback_sequence_item_grade(request):
...
+    """
+    Build a dictionary with data for callback by Web Socket.
+
+    Check flags ui_option, congratulation_message and validate score.
+    """
+    web_socket_message_dict = {}
```
idegtiarov: I would use defaultdict here.

bridge_adaptivity/module/static/module/js/forbid_next.js
AndreyLykhoman: It needs for showing this message only once time if the user has enough score to do that.

bridge_adaptivity/config/settings/base.py
```diff
@@ -205,3 +205,5 @@
         },
     },
 }
+# Use for checking score and show congratulation message or not.
+CONGRATULATION_SCORE_LEVEL = 0.95
```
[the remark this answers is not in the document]
AndreyLykhoman: I wrote a question about it to Colin.

bridge_adaptivity/module/templates/module/modals/congratulation_message_modal.html
AndreyLykhoman: I wrote the question about it ti Colin.

bridge_adaptivity/module/views.py
AndreyLykhoman: It seems no do some positive effects for this code if we would use `defaultdict`. Because I use the only 'True'  value of bool type. The default value of bool is `False`. It isn't suited for this part of the code.  This code creates the dictionary with `True` values for keys `is_button_enable` and  'is_show_pop_up'. And the last isn't created if a score is low.  The `ui_details` is not a bool value. This is a dictionary.

bridge_adaptivity/module/views.py
AndreyLykhoman: Done

bridge_adaptivity/module/static/module/js/work_with_cookie.js
AndreyLykhoman: Done

bridge_adaptivity/module/static/module/js/forbid_next.js
idegtiarov: I could be cached on the backend doesn't it

bridge_adaptivity/config/settings/base.py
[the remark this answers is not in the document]
idegtiarov: If this param is hardcoded, why don't we assign it directly in the view.py

bridge_adaptivity/module/forms.py
```diff
@@ -153,9 +153,14 @@ class Meta:
...
             'ui_option',
-            'ui_next'
+            'ui_next',
+            'congratulation_message',
         )
-        labels = {'ui_option': _('UI Option'), 'ui_next': _('Additional NEXT Button')}
+        labels = {
+            'ui_option': _('UI Option'),
+            'ui_next': _('Additional NEXT Button'),
+            'congratulation_message': _('Show congratulation message if score is 95%+')
```
idegtiarov: If our congrats param is updated in the settings, the message is incorrect

bridge_adaptivity/module/views.py
```diff
@@ -675,6 +675,24 @@ class SequenceComplete(LtiSessionMixin, DetailView):
     template_name = 'module/sequence_complete.html'
 
 
+def sync_collection(request, slug, api_request=None):
```
idegtiarov: Please clarify why was this change done?

bridge_adaptivity/config/settings/base.py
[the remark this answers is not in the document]
AndreyLykhoman: This setting is used in forms.py in the CollectionOrderForm class (I added this).

bridge_adaptivity/module/static/module/js/forbid_next.js
AndreyLykhoman: We talked about it. Cookie works well with it.  In my point, this is a frontend's work to show the message or not.

bridge_adaptivity/module/static/module/js/work_with_cookie.js
AndreyLykhoman: done

AndreyLykhoman: Added

bridge_adaptivity/module/forms.py
AndreyLykhoman: I changed it. I used the settings for showing the correct level.

bridge_adaptivity/module/views.py
AndreyLykhoman: I wrong when I moved the  function `_check_and_build_web_socket_message`. I fixed it.

bridge_adaptivity/module/views.py
```diff
@@ -675,6 +675,23 @@ class SequenceComplete(LtiSessionMixin, DetailView):
     template_name = 'module/sequence_complete.html'
 
 
+def _check_and_build_web_socket_message(sequence, score):
+    """
+    Build a dictionary with data for callback by Web Socket.
+
+    Check flags ui_option, congratulation_message and validate score.
+    """
+    web_socket_message_dict = {}
```
idegtiarov: I would create dict with the default key: value pair:
```
web_socket_message_dict = {"is_button_enable": True}
```

bridge_adaptivity/module/forms.py
```diff
@@ -153,9 +154,16 @@ class Meta:
...
-            'ui_next'
+            'ui_next',
+            'congratulation_message',
         )
-        labels = {'ui_option': _('UI Option'), 'ui_next': _('Additional NEXT Button')}
+        labels = {
+            'ui_option': _('UI Option'),
+            'ui_next': _('Additional NEXT Button'),
+            'congratulation_message': _('Show congratulation message if score is {}%+').format(
+                settings.CONGRATULATION_SCORE_LEVEL * 100
```
idegtiarov: What about our experience with the math operations over float numbers? The result could be unpleasantly long.

idegtiarov: Actually, if the instructor cannot handle the congrats treashold I would avoid any certain numbers in the message. Or cast it to int type.

AndreyLykhoman: You are right! I forgot the problem related to the math operations over float numbers. I transformed it to the int type.

bridge_adaptivity/module/views.py
AndreyLykhoman: It makes sense. I have done it.

